PR #80, opened 2015-02-24
Update _core.pyx

mkassner:
I m running PyAV in a bundle using pyinstaller. 

Without this modification codecs are not found when running PyAV inside an application bundle.

I m not sure why this is not required when running PyAv as a script (the 'normal' way).


mikeboers:
Oooookay. I wrote a little comment about it too so I don't remove it again later. ;)

PR #84, opened 2015-04-11
fix for building on windows using mingw64-32

jojje:
Patch that makes it possible to compile the project on windows using the [mingw64 compiler](http://mingw-w64.yaxm.org/doku.php).
I've only tested with the 32-bit version of the compiler since I'm only using the official 32-bit version of python 2.7 on that platform.

I imaging this patch will have to be reworked a lot before you feel it could be considered, since it *smells* a bit forced atm. However I've tried to group the windows specific parts into one conditional per file, to make it easy to spot what parts are "generic" and which are very platform specific.

The patch requires the user to put the compiled libav libraries and their two mingw dll dependencies into the av-directory, so that your cython wrappers can find them at runtime. If the user forgets to do that, a "helpful" assertion error will be triggered on the windows platform.

Final remark is that since there's no standard directory for libraries on windows, people who build ffmpeg on that platform may place the code and build artifacts anywhere. Since you rely on pkg-config to figure out how which source files to analyze and which libraries to link against, windows users will have to give setuputils a hint. That is, set `PKG_CONFIG_PATH` to point to the `pkgconfig` directory that resulted from their build of ffmpeg.

The unix/linux behavior should be unchanged.

Let me know what you think.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,15 @@ default: build
 cythonize:
 	python setup.py cythonize
 
+wheel: build-mingw32
+	python setup.py bdist_wheel
+
+build-mingw32:
+	# before running, set PKG_CONFIG_PATH to the pkgconfig dir of the ffmpeg build.
+	# set PKG_CONFIG_PATH=D:\dev\3rd\media-autobuild_suite\local32\bin-video\ffmpegSHARED\lib\pkgconfig
+	CFLAGS=$(CFLAGS) LDFLAGS=$(LDFLAGS) python setup.py build_ext --inplace -c mingw32
+	mv *.pyd av
+
 build:
 	CFLAGS=$(CFLAGS) LDFLAGS=$(LDFLAGS) python setup.py build_ext --inplace --debug
 
--- a/av/__init__.py
+++ b/av/__init__.py
@@ -1,3 +1,8 @@
+# Add the native FFMPEG and MinGW libraries to executable path, so that the
+# AV pyd files can find them.
+import os
+if os.name == 'nt':
+    os.environ['PATH'] = os.path.abspath(os.path.dirname(__file__)) + os.pathsep + os.environ['PATH']
 
 # MUST import the core before anything else in order to initalize the underlying
 # library that is being wrapped.
--- a/scripts/generate_video.py
+++ b/scripts/generate_video.py
@@ -28,10 +28,16 @@ def which(program):
 ffmpeg_cmd = None
 avconv_cmd = None
 
-if which('ffmpeg'):
-    ffmpeg_cmd = [which('ffmpeg'), '-y', '-f', 'lavfi', '-i']
-if which('avconv'):
-    avconv_cmd = [which('avconv'),'-y', '-filter_complex']
+if os.name == 'nt':
+   ffmpeg_bin, avconv_bin = 'ffmpeg.exe', 'avconv.exe'
+else:
+   ffmpeg_bin, avconv_bin = 'ffmpeg', 'avconv'
+
+
+if which(ffmpeg_bin):
+    ffmpeg_cmd = [which(ffmpeg_bin), '-y', '-f', 'lavfi', '-i']
+if which(avconv_bin):
+    avconv_cmd = [which(avconv_bin),'-y', '-filter_complex']
     
 if not ffmpeg_cmd and not avconv_cmd:
     print('Unable to find ffmpeg or avconve command')
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from __future__ import print_function
 
 from distutils.core import Command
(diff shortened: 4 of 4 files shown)


mikeboers:
(I'm not in a great place at work right now to be able to massage this into place, but...)

Thank you so much for taking this problem on! I have been hating that I have no good response when people ask about Windows.

It doesn't smell that bad at all. I certainly get where it is coming from, and won't take much at all to make it all feel unified.

I haven't checked in a while, but is there a (free) Windows-based integration testing service yet? This is critical for me, since I (currently) have no regular exposure to Windows in my personal or professional life.

mkassner:
We need this too and will have a look at the commits. Specifically for X64.

tacaswell:
appveyor will do windows integration testing.  I think scikit-image and IPython both use it.

jojje:
@tacaswell Great suggestion, looks like Appveyor is free for open source projects, but that probably means Mike will have to register the account since he's the project owner..

The only other idea I had was that perhaps using a Dockerized Wine32 linux build environment might be worth considering, but then again, Wine is rather finicky as is. I'll take a look at that option when things quiet down at work in a few weeks.

PS. Also notices that one of the test runs seem to fail for the patch, the ffmpeg branch specifically. Looks to be that the ffmpeg executable simply isn't found on the path. Hopefully trivial to fix.

jojje:
Ok gentlemen, I realized I had broken *Torvald's sacred law* accidentally breaking part of the testing for non-windows environments. Don't know what I was smoking at the time, that made me not realize changing the name of ffmpeg for the which(ffmpeg) statement in the asset generation file would lead to the command not being found :o

Thanks for not behaving like Linus :)

Also realized that commits done after the pull request to _my little private branch_ seemed to propagate over to this PR, so I had to basically botch that branch and rewrite its history. As a result this PR now contains one squashed commit of the changes I originally intended to hand over to you guys for review.

I've tested the changes on windows as well as on the two Vagrant precise32 boxes that Mike provides in the project, and was building and running the tests in virtualenv using the same versions I spotted Travis was using. Basically reproducing the Travis steps line-by-line in each box to ensure the setup and test results were as identical as possible to Travis without actually running Travis locally. Then I committed. Both sets of tests now passes on Travis also.

Commit or not
-------------
We still have the issue of finding a suitable CI for windows builds, but as it stands, it seems the added code doesn't provide any functional regression for non-windows builds, and as such the addition might provide some value to windows users, and _only_ to them.

Luckily _(or sadly)_ windows users are used to being regarded as second class citizen of the open source community, so a bit of tinkering is often acceptable _as long as_ the project can be _built_ and _run_ in most cases. Until a CI is in place, a caution in the readme such as "you're pretty much on your own until the infra-situation improves" might be an idea to consider. Those who are adventurous enough and/or are faced with real needs will at least have an avenue to use this project on that platform, even in the current state.

Since I killed my readme containing the step-by-step instructions to get going on windows, to minimize the changes in this PR, I'll just outline the steps here again, so that other may find them and for you to incorporate where you deem suitable.

Build instructions for Win32
----------------------------
1. Compile ffmpeg using the mingw64 compiler with **shared libraries** enabled.
1. Make sure MinGW's GCC compiler is the first `gcc` found on the path.   
    _Important if you have Cygwin on the system as well._
1. Set the environment variable `PKG_CONFIG_PATH` to the where the pkgconf files for ffmpeg reside.  
    Example:  `set PKG_CONFIG_PATH=c:\ffmpeg_build\lib\pkgconfig`
1. Copy the following ffmpeg libraries to the project's `av` folder:  
      `avcodec-56.dll` `avdevice-56.dll` `avfilter-5.dll` `avformat-56.dll`  
      `avutil-54.dll` `postproc-53.dll` `swresample-1.dll` `swscale-3.dll`  
    Also copy the two dependent DLLs from mingw to the same folder:  
      `libgcc_s_dw2-1.dll` `libwinpthread-1.dll`
1. Build the project: `make build-mingw32`
1. Create a self contained wheel archive that you can install on any machine.  
    `make wheel`
1. Install the package: `pip install dist/av-0.2.2-cp27-none-win32.whl`

*Note: don't try to import `av` while residing in the PyAV source folder, because it contains a directory named ... `av`, and its contents will be loaded instead of your newly built package. Change to some other directory before trying to import, else python will get confused.*

tacaswell:
@jojje As a side note, github tracks _branches_ not _commits_ in PRs so anything added to the branch of the PR is automatically added to it (and if you do a force push completely changed).  This is most definitely a feature, not a bug.

mikeboers:
I agree that since there isn't going to be any regressions, I'll pull it now. I've also added you instructions to the main docs at <https://mikeboers.github.io/PyAV/#windows>.

Now to get AppVeyor to behave...

PR #86, opened 2015-04-30
Fix: AudioStream._codec_context.coded_frame can be None, causing SEGV

radek-senfeld:
Hi, I ran into this issue when trying to encode files with audio track. Took the same approach as you did in VideoStream.encode.

I'm wondering if coded_frame will ever be set in case of audio frames. Seems like coded_frame is referenced by video/image codecs only..

http://ffmpeg.org/doxygen/trunk/structAVCodecContext.html#afdebc347b4f74e0b9271ff37cabc96e8

### DIFF
--- a/av/audio/stream.pyx
+++ b/av/audio/stream.pyx
@@ -148,8 +148,9 @@ cdef class AudioStream(Stream):
                 self._stream.time_base
             )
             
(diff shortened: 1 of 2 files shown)


mikeboers:
Merged!

Thanks for your patience. =]

PR #88, opened 2015-05-06
Conda packages for PyAV, and faster Travis builds

danielballan:
This PR makes Travis builds about 15X faster. It installs the dependencies, including FFmpeg, from binary conda packages instead of source.

The packages are downloaded from a new [community conda channel](https://binstar.org/community) where I uploaded FFmpeg, x264, and pkg-config. (We need a conda-installed pkg-config, not system pkg-config, to see the libraries of conda-installed FFmpeg.)

We can also upload packages for PyAV itself. I put some on my personal conda channel to try. @tacaswell and I tested this on OSX and 64-bit Linux. There are packages for Python 2.7, 3.3, and 3.4.

```
conda install -c danielballan PyAV
```

That one line takes care of everything. The installation instructions could get a lot simpler!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+sudo: false  # faster, container-based builds
 
 python:
     - "2.6"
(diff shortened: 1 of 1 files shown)


mikeboers:
I merged everything up to using conda for Travis (even though that is the point of the title). I'd be happy to test a conda build in addition to the others, but I need the source build for the foreseeable future.

Please convince me that I don't need to. =]

felixlaumon:
Is there any updated conda channel for PyAV. The one from `conda install -c danielballan PyAV` is on v0.2.3.

mikeboers:
I think I'm happy to declare with certainly that I want to keep testing the source on Travis (with the understanding that the conda-forge setup is testing itself (a little)).

Issue #105, opened 2015-07-10
Libraries not found on Ubuntu 14.04


happycharleswang:
Hi!
I was installing  on ubuntu LTS 14.04,         sudo python setup.py install

errors came out : 

~~~
Could not find avformat with ctypes; looked in:
    libavformat.so.54
    /usr/local/lib/libavformat.so.54
Could not find avutil with ctypes; looked in:
    libavutil.so.52
    /usr/local/lib/libavutil.so.52
Could not find avcodec with ctypes; looked in:
    libavcodec.so.54
    /usr/local/lib/libavcodec.so.54
Could not find ['avformat', 'avutil', 'avcodec'] with ctypes.util.find_library
Some libraries can not be found for inspection; aborting!
~~~

I have looked in previous issues post here,  they are similar.
You said to remove some thing and then compile, but that thing is already removed in the current  version, right?
So why this problem is still there.... I am confused... someone please help, will appreciate.

mikeboers:
I'm sorry to say there isn't enough information here for me to help. Can you tell us anything about your library installs, what pkg-config returns, and any investigations into the paths listed in the error?

With a clean Ubuntu 14 the complete build in a virtualenv should be very similar to:

~~~
sudo apt-get install -y python-dev python-virtualenv pkg-config
sudo sudo apt-get install -y \
    libavformat-dev libavcodec-dev libavdevice-dev \
    libavutil-dev libswscale-dev libavresample-dev
git clone git@github.com:mikeboers/PyAV.git
cd PyAV
virtualenv venv
. venv/bin/activate
pip install Cython
python setup.py build
~~~

happycharleswang:
hi Mike, thx for reply!

It seems ctypes was trying to find files in /usr/local/lib
but my files are located in another folder:
/usr/lib/x86_64-linux-gnu/libavcodec.so.54
/usr/lib/x86_64-linux-gnu/libavformat.so.54
/usr/lib/x86_64-linux-gnu/libavutil.so.52

yet
static libs are located in /usr/local/lib
/usr/local/lib/libavcodec.a
/usr/local/lib/libavformat.a
/usr/local/lib/libavutil.a



            pkg-config --libs libavcodec
-pthread -L/usr/local/lib -lavcodec -lavresample -lavutil -lm 
 
             python -c 'import ctypes.util; print ctypes.util.find_library("avcodec")'
libavcodec.so.54

any clue?
thanks!

happycharleswang:
[...]

Issue #52, opened 2014-09-17
Deprecated struct members in unstable Libav builds (e.g. AVStream.r_frame_rate)


michaeljoelphillips:
Greetings,

I've attempted to install PyAV with pip as well as compile it from source and I get the following errors.

```
building 'av.video.stream' extension

gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -DHAVE_AVFORMAT_CLOSE_INPUT=1 -Iinclude -I/usr/include/python2.7 -Ibuild/temp.linux-x86_64-2.7/include -c /home/nomad/build/av/src/av/video/stream.c -o build/temp.linux-x86_64-2.7/home/nomad/build/av/src/av/video/stream.o

/home/nomad/build/av/src/av/video/stream.c: In function ‘__pyx_pf_2av_5video_6stream_11VideoStream_12guessed_rate___get__’:

/home/nomad/build/av/src/av/video/stream.c:2549:90: error: ‘struct AVStream’ has no member named ‘r_frame_rate’

error: command 'gcc' failed with exit status 1
```

```
Downloading/unpacking av
  Running setup.py egg_info for package av
    Could not find ['avcodec', 'avutil', 'avcodec'] with ctypes.util.find_library
    Could not find ['avformat'] with ctypes.util.find_library
    Could not find ['avutil'] with ctypes.util.find_library
```

I'm assuming I'm missing a dependency or something, but to the best of my knowledge, I have all of the correct development packages for the dependencies listed in the wiki.  Any and all help is appreciated.

PR #106, opened 2015-07-13
Use distutils.ccompiler to test for functions and structure members

mikeboers:
This removes the use of `ctypes` to discover what functions are available. Instead, it uses the same C compiler that Python would end up using to compile the modules in order to build and link tiny test executables and discover if those functions exist.

This also allows for detecting members of structures, so we can safely implement deprecated members.

This breaks our Window's build (temporarily, and only because I can't maintain it).

Any thoughts?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -251,7 +251,7 @@ class ConfigCommand(Command):
         )
 
         for ext in self.distribution.ext_modules:
-            for key, value in extension_extra.iteritems():
+            for key, value in extension_extra.items():
                 setattr(ext, key, value)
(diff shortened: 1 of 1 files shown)


mikeboers:
The builds passed; that's almost good enough for me...

mikeboers:
If we merge this, we can reconsider #52

mikeboers:
Using this would likely solve the various static-vs-dynamic problems we have had, e.g. #105 (and at least one more issue that I don't care to dig out).

PR #109, opened 2015-07-27
pyio seek's return type should be int64_t

xxr3376:
PyAV will throw OverflowError: value too large to convert to int when opening a python file object larger than 2GB, but open by filename is ok.
I found that pyio_seek_gil should return int64_t instead of int for opening a large file.

Sorry for wrong insert about setup.py, this pull request has reseted that commit.


mikeboers:
Great! Thanks for catching that. It is now merged.

PR #114, opened 2015-08-12
Fix bug, and remove workaround in setup.py

vidartf:
Previously, `mod_name` in setup.py was improperly escaped. This only became apparent on Windows, where a workaround had been implemented to circumvent the effects. This PR fixes the bug, and removes the workaround (which became a bug once the initial one was fixed).


mikeboers:
I've fixed this to work with non-Windows, and it will be pushed/merged in the near future. Thanks!

PR #115, opened 2015-08-12
Improved setup + platform independence

vidartf:
Fixes many smaller issues, and introduces changes that I hope should be straight-forward to accept.


mikeboers:
Thanks for what appears to be a considerable effort, and significant improvement to the build system!

I'm going to have to go commit-by-commit through this, since I see a few errors (not freeing a new pointer, using a `log` that doesn't exist, not accounting for `ImportError` on non-Windows, etc.), and I'm going to pass on the automatic PEP8 conformity pass.

Cheers!

mikeboers:
I went commit by commit cherry-picking and fixed a few things here and there (I was wrong about the `log`). All merged now.

Thanks again!

PR #116, opened 2015-08-13
Improved windows setup

vidartf:
This PR includes several features to fix compiling with MSVC, and several features to fix compiling without pkg-config.

Thanks for taking the effort to merge the previous PRs nicely! This is my final "major" PR for this. With this one, everything works nicely for me with both MinGW and MSVC with the official ffmpeg distros. I'll try to write up an updated install/build instruction when you've had the chance to review/merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,14 @@
 /docs/includes.rst
 /dist
 *.egg-info
+/av/**/*.exp
+/av/**/*.lib
+/av/**/*.pdb
+/msvc-projects
+/*.sln
+/*.suo
+/*.sdf
+/ipch
 
 # Testing.
 /sandbox
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 # Build products.
 /src
 /build
+/av/**/*.pyd
 *.c
 *.so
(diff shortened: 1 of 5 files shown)


vidartf:
For now, this is a typically build command for me: 
`> python setup.py clean --all build_ext --inplace --ffmpeg-dir=J:\prog\ffmpeg -c msvc`

mikeboers:
Wonderful!

What would be absolutely amazing is if you could assist in getting the build to work on appveyor. It would actually give me continuous feedback about the Windows build, and put it within the realm of possibility that I won't break it (again).

vidartf:
Well, I don't have much knowledge of these setups, but for MSVC build support, you'd have to pull the MSI from this page http://www.microsoft.com/en-us/download/details.aspx?id=44266.
You'll also need two headers from here (could possibly be in repository): https://code.google.com/p/msinttypes/
I'm not sure if python is already installed on the VM, but if not I'd recommend WinPython as it comes prebundled with a lot of useful things, e.g. mingw.

mikeboers:
Be warned that I likely won't be able to visit this until early next week... but I will.

PR #135, opened 2015-10-15
Adding Data field in packet

anshul1912:
Signed-off-by: Anshul Maheshwari <er.anshul.maheshwari@gmail.com>


mikeboers:
Thanks for the contribution! However, I'd like to use the buffer protocol for data access. I just pushed 9f736b5aa4d8 which provides it.

anshul1912:
how to use it?
can you please provide me some example to get data out of it.
should I call pkt.buffer_ptr  or pkt.to_byte()

anshul1912:
Ok, I got it, to use it I have to call pkt.to_byte().

Thanks for your great work

PR #144, opened 2015-11-27
Add GENPTS flag when create context

xxr3376:
I got None when I'm reading a h.264 file, I found that I can force LibAV calculate PTS by setting GENPTS flag.

This flag may lead to extra read when calculating PTS, may be we should leave this flag as a option when opening a file?


mikeboers:
This looks like it could be a great idea. Give me a bit of time to do some reading.

Issue #148, opened 2016-02-17
Random audio channel layout when there are more than 8 channels of audio


tahpee:
I've got a video clip that has 16 channels of audio. When I try to decode the audio packets I get a random channel layout. It turns out that it is looking up an entry in default_layouts (audio/layout.pyx) for 16 channels and there is no entry, so it reads some random rubbish from memory and tries to use that as the channel layout.

Oddly the same file and script is fine on Ubuntu with PyAV 0.2.4 and LibAV but under OS X with ffmpeg 2.8.6 it fails.

Fix is to add some entries in default_layouts for channel counts > 8.

jlaine:
What's the status of this? It looks like the corresponding PR was merged?

mikeboers:
Looks like.

PR #149, opened 2016-02-17
Added simplified entries for channel counts greater than 8

tahpee:
Fix for #148

### DIFF
--- a/av/audio/layout.pyx
+++ b/av/audio/layout.pyx
@@ -15,7 +15,7 @@ cdef AudioLayout get_audio_layout(int channels, uint64_t c_layout):
 
 
 # These are the defaults given by FFmpeg; Libav is different.
-cdef uint64_t default_layouts[9]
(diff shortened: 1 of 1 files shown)


mikeboers:
Thank you for digging into this! I'll give this attention as soon as I can. It might take a few days.

Cheers,
Mike

mikeboers:
Is there some documentation that refers to how this is calculated? I foresee myself wondering WTF this means when reading it in the future.

tahpee:
To be honest, I just kept turning on bits from the right until I had 16 channels worth. It looks like libav 12 includes support for more than 16 channels of audio. 
https://git.libav.org/?p=libav.git;a=blob;f=libavutil/channel_layout.h;h=5bd0c2ce6ea0d7a88ba30d221eeb76ecd1c2ef04;hb=ee480790c7eeb03c9cebd8971c46e0cb7db65277

PR #189, opened 2016-09-18
Add Encoder and Decoder classes

markreidvfx:
The following pull request adds new `Encoder` and `Decoder` classes. This allows one to setup a encoder/decoder without a AVFormatContext. Currently its a little verbose to setup, but that can be made easier in future patches.

here is a quick example that generates mpeg4 packets:

``` python
import av
# setup encoder
encoder = av.Encoder("mpeg4")
encoder.width = 640
encoder.height = 480
encoder.time_base = "25/1"
encoder.pix_fmt = "yuv420p"

# create a blank frame
frame = av.VideoFrame(640, 480, 'yuv420p')

# encode 10 frame
for x in range(10):
    for packet in encoder.encode(frame):
        print packet

# flush encoder
for packet in encoder.flush():
    print packet
```

The new classes are also handy for reading and writing image formats. We use could them to  add read and write methods on the `VideoFrame` object that supports all the formats that libavcodec supports.  There are some added tests that show how to do this.

The last patch in the series integrates the new class in the Stream object instead of directly calling libavcodec. As a first step, I tried to do it in a way did that doesn't break anything. There is currently a lot of shared functionality currently between the `Stream` object and the `CodecContext` object. I like to clean that up a bit more, but again this is just the first step.

### DIFF
--- a/av/__init__.py
+++ b/av/__init__.py
@@ -21,3 +21,5 @@ from av.container import open
 from av.utils import AVError
 from av.video.format import VideoFormat
 from av.video.frame import VideoFrame
+from av.encode import Encoder
+from av.decode import Decoder
--- a/av/audio/fifo.pyx
+++ b/av/audio/fifo.pyx
@@ -114,7 +114,13 @@ cdef class AudioFifo:
             self.pts_offset -= nb_samples
         
         return frame
-    
+
+    def iter(self, unsigned int nb_samples=0, bint partial=False):
+        frame = self.read(nb_samples, partial)
+        while frame:
+            yield frame
+            frame = self.read(nb_samples, partial)
+
     property samples:
         """Number of audio samples (per channel) """
         def __get__(self):
--- a/av/audio/fifo.pyx
+++ b/av/audio/fifo.pyx
@@ -85,6 +85,9 @@ cdef class AudioFifo:
         if not partial and self.samples < nb_samples:
             return
 
+        if partial:
+            nb_samples = min(self.samples, nb_samples)
+
         cdef int ret
         cdef int linesize
         cdef int sample_size
--- a/av/audio/stream.pxd
+++ b/av/audio/stream.pxd
@@ -12,14 +12,6 @@ cdef class AudioStream(Stream):
 
     cdef readonly AudioLayout layout
     cdef readonly AudioFormat format
-    
-    # Hold onto the frames that we will decode until we have a full one.
-    cdef AudioFrame next_frame
 
     # For encoding.
     cdef AudioResampler resampler
-    cdef AudioFifo fifo
-
-    cpdef encode(self, AudioFrame)
-
-
--- a/av/audio/stream.pyx
+++ b/av/audio/stream.pyx
@@ -43,25 +43,8 @@ cdef class AudioStream(Stream):
     property channels:
         def __get__(self):
             return self._codec_context.channels
-        
-    cdef _decode_one(self, lib.AVPacket *packet, int *data_consumed):
 
-        if not self.next_frame:
-            self.next_frame = alloc_audio_frame()
-
-        cdef int completed_frame = 0
-        data_consumed[0] = err_check(lib.avcodec_decode_audio4(self._codec_context, self.next_frame.ptr, &completed_frame, packet))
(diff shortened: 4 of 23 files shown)


mikeboers:
Whoa! Thanks for making that happen!

I was working on something similar in codec-ctx branch, but it seems like you got a lot further. I did it in a slightly more typical PyAV style, which certainly doesn't mean it is better by any means, but I think I caught a bunch of edge cases that you did not. Gotta pick one to adapt into the other, I guess.

I'm hoping to find time to make a PyAV sprint soon so I can reconcile this, and the new Windows fixes, and etc..

markreidvfx:
ya, sorry for the surprise!  I needed a standalone encoder for something I'm working on.
I'll take a closer look at your branch. 

A first glance, I can see you where setting up a `AVCodecParserContext` which more of a decoding thing, I was mostly focusing on encoding, sound like a really nice thing to have on a decoder.   Also moving all this to a av/codec/ sub module is something I was gonna do too. I can do some more work on it to get it more PyAV style if you want (just don't know what that means :p ).

if you do hack a couple of things I noted where:
`avcodec_copy_context` is really unreliable (and deprecated, they have this new AVCodecParameters stuff now). I did try using it to duplicate the CodecContext on a stream created from a FormatContext, so I wouldn't have to worry about GC as much, but I couldn't get it working reliably. might have been doing something wrong, but gave up.

The streams, CodeContexts, etc created and added to a AVFormatContext are owned by it. they are freed when the formatContext closes (though, CodeContexts need to be manually close before that). In standalone mode freeing the CodeContext is simple (close and free), but in attached to a stream mode they only need to close. its a little tricky to get the GC correct, but what I did I think is correct, holding a ref to the ContainerProxy.

tjcrone:
The changes in @markreidvfx's fork adds really helpful functionality, making it possible to decode packet payload without having a container object. It would be great to get this merged into the main fork if possible.

markreidvfx:
I now have this pull request rebased on top of the current head in a branch on my fork. I was thinking of changing a couple more things, but parts of it I'll try and send separately.

I was slowly looking into using the new encode/decoder api
https://www.ffmpeg.org/doxygen/3.1/group__lavc__encdec.html
but am not too sure about compatibility.
There is also this CodecParameter struct thingy there changing everything to use...

mikeboers:
Hey all.

I've just pushed (to master) my first version of a mashup of @markreidvfx's work and my own. The API is a little different, but not terribly.

Thanks everyone for the hard work and patience!

I'll leave this open until I verify Travis is passing, and I finish the most basic of documentation.

mikeboers:
I've rolled back master (I know, I know) since my version is obviously incomplete, now that I've been trying to run @markreidvfx's tests (which wonderfully point out that I'm dropping a frame).

Mark, is there a particular reason you switched to `encode(...)` returning an iterator instead of one packet? I've been feeling like that is the thing to do for a while (since I'm paranoid that there may be multiple packets for each input frame in some container/codec combinations), but I don't have much proof.

markreidvfx:
I think the reason I made `encode(...)` a generator was to try and unify it with flushing.

tjcrone:
Don't you need multiple frames to encode packets when each packet does not represent a keyframe?

mikeboers:
@tjcrone Yeah. PyAV up to this point assumed the relationship was one packet to multiple frames. I have, however, been uneasy about the inverse relationship for a long time, so I've been considering making both directions return lists of their return values.

I've decided to do it, and maybe try to maintain backwards compatibility.

Regardless, at this point, I'm pretty sure I've essentially merged the functionality that @markreidvfx did with what I was working on. Having both made it a ton easier to get this done. Thank you, many more times!

I'm still failing some tests on Travis, but I'm considering merging to master at this point.

markreidvfx:
great stuff @mikeboers, thanks for looking further into this.

mikeboers:
I just pushed it to master (again). All the tests pass, but I fear I might have broken a small feature here or there that was untested.

Frankly, I just spent way too much time doing that, and need it out of my life for the week. So onto master it goes so that hopefully someone will tell me what I did wrong. ;)

PR #192, opened 2016-12-10
Add an example of writing a movie from numpy

videan42:
I'm trying to start using PyAV to replace a bunch of ugly shell calls to ffmpeg. It's working great so far, but I've mostly been learning the library by reading your test cases.

I was wondering if some of the simpler ones could be made into an ``examples`` docs page. This pull is for ``tests/test_encoding.py`` which is a good example of how to write a movie from numpy/PIL.

I'd be happy to add a few more if you think this is a good idea.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -1,0 +1,45 @@
+Examples
+========
+
+Writing a movie from numpy
+
+::
+
+    duration = 10  # Seconds
+    frames_per_second = 24
+    total_frames = duration * frames_per_second
+
+    container = av.open('test.mp4', mode='w')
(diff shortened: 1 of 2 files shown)


mikeboers:
Hey David,

Thanks for taking the initiative to do this. I've been wanting to get some real examples into the docs for a **long** time.

I didn't reply immediately because I wanted to think about how to have the examples in the docs tested (to assert that API changes to break the docs), but I haven't had the time, and wanted to let you know how much I appreciate that you're thinking of this.

I'll pull this now, and open a new ticket.

PR #203, opened 2017-03-10
We don't need to host the conda-recipe anymore due to conda-forge.

mikeboers:
@danielballan @tacaswell Am I correct to do this?


danielballan:
IMO, yes. Now that we have conda-forge, that's a better place to keep the recipe. This one should be removed so we don't have to remember to keep it in sync.

mikeboers:
Good enough for me. Merged via rebase.

PR #205, opened 2017-03-11
Rebase Appveyor build on conda

caspervdw:
@mikeboers I noticed the Appveyor build to fail for no apparent reason. To save us from corrupt downloads, I changed the test recipe, pulling in `ffmpeg` and `msinttypes` binaries from conda-forge. This does limit the ffmpeg version to 2.8.6, as that is the only version currently on conda.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -72,7 +72,7 @@ install:
   - ps: "ls \"C:/Program Files/Microsoft SDKs/Windows\""
 
   # Prepend to the PATH of this build
-  - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+  - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%CONDA_ROOT%\\Library\\bin;%PATH%"
 
   # Install dependencies
   - conda config --set always_yes true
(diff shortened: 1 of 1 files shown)


mikeboers:
Taking my opinion with a grain of salt because I've never run this on Windows, and never really read our own Windows build system...

While not ideal, it sounds good enough to me, and is likely to make the Windows testing more meaningful if it doesn't spontaneously fail on us. Linux/macOS test our compatibility with various FFmpeg/LibAV versions, and that should catch 99% of the differences there... right?

You added a couple more commits to this; let me know when to hit the "pull" button. 😉

caspervdw:
This apparently doesn't fix anything, but still I am :+1: on using conda to pull in ffmpeg and msinttypes as it is more reliable this way.

For some obscure reason, the Py34 tests keep failing, while nothing has changed between https://ci.appveyor.com/project/mikeboers/pyav/build/1.0.211 and https://ci.appveyor.com/project/mikeboers/pyav/build/1.0.210.

For me the idea of the windows builds is checking if it compiles succesfully using the various MSVC compilers (Py27, Py34, Py35 on both 32 and 64 bit) The rest is indeed covered by the Linux builds. I am afraid that the failing Py34 build will come back at us at conda-forge after the v0.3.3 release. But maybe it will solve itself...

PR #217, opened 2017-05-04
Some Bug fixes and add Packet __init__

markreidvfx:


mikeboers:
I've pulled in the two small fixes. The Packet init is being handled by the new codec context patch. I'll go discuss that in the other PR.

🍻

PR #224, opened 2017-05-09
Update appveyor ffmpeg version

caspervdw:
A newer FFMPEG version is available from conda-forge, so that we can easily update the tests. 
I additionally removed Python 3.3 and added 3.6 to the tests.


mikeboers:
Thanks! Pulled.

PR #230, opened 2017-05-24
avcodec_receive_*(): EAGAIN can be -35, but also -11

crackwitz:
I've got ffmpeg 3.3.1 on win7 x64 and cpython 3.6.1 and this caused trouble. renaming the constants to something obvious might be a good idea, but this got it working for me at least.


mikeboers:
Thanks for that. I cherry-picked this, and then replaced it with the `libc.errno` version that Cython provides.

Cheers,

Mike

Issue #233, opened 2017-05-29
OutputContainer.mux() puts all packets with stream_index=0


vbkaisetsu:
I tried to write a small encoder using the latest git repository for a movie that contains a video and an audio.
However, both kinds of packets are mixed in to stream index "0".

OutputContainer.mux_one refers packet.struct.stream_index, but it does not set in Stream.encode. It may be removed on ae9801267c55d719c2cd9d8a9e4877d764766f3f.

mikeboers:
I just fixed this with 3dd212a0cbe0c74f1ff765dbce4b99bdf510abe1.

🍻

PR #234, opened 2017-05-29
Set stream info for packets

vbkaisetsu:
Fix a bug #233


mikeboers:
I just fixed this with 3dd212a0cbe0c74f1ff765dbce4b99bdf510abe1. Thanks for your help pointing out the exact problem.

PR #240, opened 2017-06-12
Manage Data Stream

mgoacolou:
This PR give ability to manage data stream 

there is an  issue with stream because there is not codec for this kind of data so the properties "codec", "type" and name must be defined.

### DIFF
--- a/av/container/streams.pxd
+++ b/av/container/streams.pxd
@@ -9,6 +9,7 @@ cdef class StreamContainer(object):
     cdef readonly tuple video
     cdef readonly tuple audio
     cdef readonly tuple subtitles
+    cdef readonly tuple data
     cdef readonly tuple other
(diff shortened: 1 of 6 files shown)


mikeboers:
Thanks for this.

I'm curious, what are you able to do with this except know that the data stream exists? I suppose then you can identify packets and look at the raw data...

mgoacolou:
With this, we are able to read Full Motion Video/ MISB video and extract geo information (center of image, corners, ...) of frames

mgoacolou:
for example: https://community.esri.com/docs/DOC-8527

mgoacolou:
Hi,
Do you plan to merge this Pull Request.

mikeboers:
Yes, when I have the time.

PR #246, opened 2017-06-25
more filter info in API

skirsdeda:
Included all filter flags for completeness and ```process_command``` pointer to check if filter supports commands. Also a small fix for ```options``` property (some generic filters don't have any descriptor). ```process_command``` is available since before ffmpeg 1.0 so it should be fine to add. Basically, now we can have all the same info about filters as provided by ffmpeg command.

### DIFF
--- a/av/codec/__init__.py
+++ b/av/codec/__init__.py
@@ -1,2 +1,2 @@
-from .codec import Codec, codecs_available
+from .codec import Codec, codecs_available, codec_descriptor
 from .context import CodecContext
--- a/av/codec/codec.pyx
+++ b/av/codec/codec.pyx
@@ -184,12 +184,14 @@ cdef class Codec(object):
         def __get__(self): return flag_in_bitfield(self.desc.props, lib.AV_CODEC_PROP_TEXT_SUB)
 
 
-
 codecs_available = set()
 cdef lib.AVCodec *ptr = lib.av_codec_next(NULL)
 while ptr:
     codecs_available.add(ptr.name)
     ptr = lib.av_codec_next(ptr)
+
+
+codec_descriptor = wrap_avclass(lib.avcodec_get_class())
 
 
 def dump_codecs():
--- a/av/codec/codec.pyx
+++ b/av/codec/codec.pyx
@@ -4,7 +4,7 @@ from warnings import warn
 
 from av.audio.format cimport get_audio_format
 from av.descriptor cimport wrap_avclass
-from av.utils cimport media_type_to_string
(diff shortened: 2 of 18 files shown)


mikeboers:
Cool, thanks!

It is going to take me a few days to look at this one, so sit tight.

skirsdeda:
Also fixed adding multiple instances of the same filter to filter graph.

skirsdeda:
Couldn't resist with one more improvement. Now you get ```OptionChoice.is_default``` (which is ```bool``` of course). This let's you quickly determine which flags are selected by default or which choice is the default one. For example gif encoder's sole option 'gifflags' has two choices which are both on by default:
```python
from av.codec import Codec
assert all(c.is_default for c in Codec('gif', 'w').descriptor.options[0].choices)
```

skirsdeda:
Not sure about the last one because ```AVClass``` used to build ```CodecContext.descriptor``` is static. So maybe a module-level variable would be better (smth like 'codec_context_descriptor'). The pointer can also be obtained by calling ```avcodec_get_class``` which could be used for module-level Descriptor instance.

skirsdeda:
Played a bit with enum support in Cython. It's really great and is just the right thing to use for all flag enumerations. Some things I discovered while playing with enums:
* constants from libav have to be defined with `cdef enum` in includes to be usable when assigning values to "cpdef'ed" enum members. Otherwise Cython complains that they're not constants. And coincidentally, this is less verbose while having the same effect (checked built .c files to make sure).
* it's better to define enum members as uppercase, otherwise we hit a lot of clashes with keywords, also uppercase members is the recommended way as seen in Python docs.
* `cpdef enum` exposes PEP-435 like enum which works the same as Python's `enum.IntFlag` so bitwise operations work like expected.

So far I only changed filter and option related code but there are many other places which would benefit from this refactoring.

mikeboers:
I like the idea of the enum stuff!

I'm still looking at a pretty big backlog in meatspace right now, so thank you for being patient with me.

🍻

skirsdeda:
If enum stuff seems good, I can at least fix the tests I broke :)

mikeboers:
Sorry for being negligent.

As far as you're concerned, the only remaining issue is the one test that fails (that I just corrected to use the new API)?

mikeboers:
Nothing. It's all me.

For some reason I thought I might be waiting on you, but I guess I read it wrong. Sorry.

mikeboers:
It has (finally) been merged. Hooray! 🎉

PR #250, opened 2017-07-09
Support specifying metadata character encoding

### LINKED ISSUE #59: Problems with international characters in metadata
I've built PyAV for python 3, and it works fine, except that it fails on reading metadata from files, if they contain unicode/international (I'm not sure) characters. I.e. I had file with title for audio channel with russian letters, and got exception:

> "UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 0: ordinal not in range(128)"

I fixed that by changing the avdict_to_dict so that it returns bytebuffers instead of strings as values:

``` python
cdef dict avdict_to_dict(lib.AVDictionary *input):

    cdef lib.AVDictionaryEntry *element = NULL
    cdef dict output = {}
    cdef bytes value
    while True:
        element = lib.av_dict_get(input, "", element, lib.AV_DICT_IGNORE_SUFFIX)
        if element == NULL:
            break
        value = element.value
        output[element.key] = value
    return output
```

Maybe you'll find some other, more nicer way, but I'm not sure if there is a way to determine right encoding.

mkfmnn:
For writing, and for reading on Python 3, the default codec changes
from ASCII to UTF-8. In Python 2, the default behavior remains to not
perform decoding on reading, but this can be overridden by explicitly
specifying an encoding. This change applies to both container and
stream metadata.

fixes #59

### DIFF
--- a/av/container/core.pxd
+++ b/av/container/core.pxd
@@ -17,6 +17,8 @@ cdef class ContainerProxy(object):
     cdef flush_buffers(self)
 
     cdef str name
+    cdef str encoding
+    cdef str errors
 
     # File-like source.
     cdef object file
@@ -50,3 +52,6 @@ cdef class Container(object):
 
     cdef readonly StreamContainer streams
     cdef readonly dict metadata
+
+    cdef str encoding
+    cdef str errors
--- a/av/container/core.pyx
+++ b/av/container/core.pyx
@@ -32,6 +32,8 @@ cdef class ContainerProxy(object):
 
         # Copy key attributes.
         self.name = container.name
+        self.encoding = container.encoding
+        self.errors = container.errors
         self.file = container.file
         self.writeable = container.writeable
 
@@ -169,7 +171,7 @@ cdef class ContainerProxy(object):
 
 cdef class Container(object):
 
-    def __cinit__(self, sentinel, file_, format_name, options):
+    def __cinit__(self, sentinel, file_, format_name, options, encoding, errors):
 
         if sentinel is not _cinit_sentinel:
             raise RuntimeError('cannot construct base Container')
@@ -189,6 +191,9 @@ cdef class Container(object):
 
         self.options = Dictionary(**(options or {}))
 
+        self.encoding = encoding
+        self.errors = errors
+
         self.proxy = ContainerProxy(_cinit_sentinel, self)
 
         if format_name is None:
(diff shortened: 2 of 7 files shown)


mkfmnn:
So far I've only manually tested this, for opening files for reading on Python 2 and 3. I was planning to take a stab at writing some automated tests but haven't really looked into what that entails yet. I wanted to at least send out a draft for feedback

mikeboers:
Thank you for taking a stab at this.

Two bits of feedback:

1. I think I like your suggestion on #59 that the kwargs be prefixed with `metadata_` so it is pretty clear what is happening.

2. My gut says that Cython has a way of magically dealing with the Py2/3 differences when it comes to the API. I remember doing something in another project, but I just don't remember what/where...

I'm happy to help write some tests if you know of some samples (ideally from the FATE suite). In this case though we can likely generate them.

mikeboers:
I've added `metadata_` prefixes to many of the attributes, and merged it.

Thanks!

mkfmnn:
Thanks for the merge! Unfortunately, one of the changes in 2c0fabc01ec2ddd9f801d4ff8bc8d36407d06d66 broke the handling of encodings for Streams: `metadata_encoding` and `metadata_errors` [need to be populated before the ContainerProxy is instantiated](https://github.com/mikeboers/PyAV/commit/2c0fabc01ec2ddd9f801d4ff8bc8d36407d06d66#diff-f5f84daccfb1cd5a2d7e4b8506861a19L194), so that they're available to be copied to the proxy object. Currently, they're both always `None` on the proxy, which results in the error reported in #265 on Python 3 when attempting to open any file. I haven't gotten a chance to do anything since originally sending out this PR but maybe I can take a stab at automating tests this weekend.

> My gut says that Cython has a way of magically dealing with the Py2/3 differences when it comes to the API. I remember doing something in another project, but I just don't remember what/where...

There's the `c_string_type` directive, which PyAV is already using ([documented here](http://cython.readthedocs.io/en/latest/src/tutorial/strings.html#auto-encoding-and-decoding), [used here](https://github.com/mikeboers/PyAV/blob/master/setup.py#L646)), but I didn't see a way to get that to solve this issue, and I imagine the flexibility of being able to explicitly customize encoding and error-handling could come in handy for working with media files from a variety of sources anyway.

Issue #95, opened 2015-05-21
Generalize to_nd_array format understanding


mikeboers:
We should be able to support most planar formats as long as we can come up with an appropriate numpy format for the buffer.

The [current implementation](https://github.com/mikeboers/PyAV/blob/af9decdf169193d98b2a9297fc9f0437525c5281/av/video/frame.pyx#L252-L259) (hardcoding the few we have used) is inadequate.

PR #252, opened 2017-07-12
More general to_nd_array 

wecacuee:
More general to_nd_array that handles when plane.line_size != plane.width * channels, data_type.itemsize

Issue #95 

Test 
```
>>> import av, urllib, PIL
>>> urllib.urlretrieve('http://fate.ffmpeg.org/fate-suite/png1/55c99e750a5fd6_50314226.png', '/tmp/000.png')
>>> frames = (p.decode_one() for p in av.open("/tmp/%03d.png").demux(video=(0,)))
>>> ndarrays = (f.to_nd_array() for f in frames if f is not None)
>>> np.all(np.array(PIL.Image.open("/tmp/000.png")) == next(ndarrays))
True
```

### DIFF
--- a/av/video/frame.pyx
+++ b/av/video/frame.pyx
@@ -248,15 +248,27 @@ cdef class VideoFrame(Frame):
             raise ValueError('Cannot conveniently get numpy array from multiplane frame')
 
         import numpy as np
+        from numpy.lib.stride_tricks import as_strided
 
         # We only suppose this convenience for a few types.
         # TODO: Make this more general (if we can)
+        plane = frame.planes[0]
         if frame.format.name in ('rgb24', 'bgr24'):
+            data_type = np.dtype(np.uint8)
+            channels = 3
+            arrshape =  (plane.height, plane.width, channels)
+            arrstrides = (plane.line_size, channels, data_type.itemsize)
             return np.frombuffer(frame.planes[0], np.uint8).reshape(frame.height, frame.width, -1)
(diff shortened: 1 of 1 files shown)


mikeboers:
Wonderful!

Do you happen to know of some test material (ideally in the FATE suite) that we can use for writing tests for this with?

wecacuee:
Yes this should work with png stream (sequence of numbered pngs) from fate suite. I included some test code in the first comment.

mikeboers:
You certainly did. I think I'm too tired for this. ;)

mikeboers:
Another general improvement: implement this in the `__getbuffer__` method, and then the `as_ndarray` becomes `np.array(self)`.

mikeboers:
Just a heads up that I'm still thinking about this periodically.

At this point I strongly want to go in the direction of having the buffer protocol also be shaped/strided, but I'm not 100% on what that implies. I want to read/experiment more in that direction.

mikeboers:
Hey @wecacuee, over here `http://fate.ffmpeg.org/fate-suite/png1/55c99e750a5fd6_50314226.png` is coming up as `rgba`, and has the same linesize as width * itemsize (1320).

Is it different for you?

wecacuee:
Sorry my bad. 

I was doing some sort of preprocessing (may be converting to ppm) before testing. I didn't test the exact code that I pasted. This time I added two test cases for 'png' and 'ppm' format.

PR #253, opened 2017-07-13
Fix SegFault resulting from incorrect FFMPEG structures deallocation

heni:
[=] fix segfault resulting from incorrect FFMPEG structures deallocation
[=] fix some tests to be runnable under python3
[+] add test for correct (segfault free) opening of file with random data
[=] fix ContainerProxy.seek behaviour for integers more then 2**31 with 32bit CPython

### DIFF
--- a/av/container/core.pxd
+++ b/av/container/core.pxd
@@ -31,6 +31,7 @@ cdef class ContainerProxy(object):
     cdef unsigned char *buffer
     cdef long pos
     cdef bint pos_is_valid
+    cdef bint is_opened
 
     cdef int err_check(self, int value) except -1
 
--- a/av/container/core.pyx
+++ b/av/container/core.pyx
@@ -128,7 +128,7 @@ cdef class ContainerProxy(object):
 
         # We used to take floats here and assume they were in seconds. This
         # was super confusing, so lets go in the complete opposite direction.
-        if not isinstance(offset, int):
+        if not isinstance(offset, (int, long)):
             raise TypeError('Container.seek only accepts integer offset.', type(offset))
         cdef int64_t c_offset = offset
 
--- a/av/container/core.pyx
+++ b/av/container/core.pyx
@@ -25,6 +25,7 @@ cdef class ContainerProxy(object):
 
     def __init__(self, sentinel, Container container):
 
+        self.is_opened = False
         cdef int res
 
         if sentinel is not _cinit_sentinel:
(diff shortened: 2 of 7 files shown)


heni:
PING!

mikeboers:
PONG! (a.k.a. I see you, and I appreciate your efforts.)

mikeboers:
So, the whole deallocation thing. What I had was kinda organic grown out of what worked for me as I went. I didn't really have a good justification for how I was doing it.

You know what would be awesome? Good justification. If you can write enough in there about why this is the one true way to deallocate that stuff, I would be extremely grateful.

Or you can explain a bit here, and I'll do it with a new understanding. :)

heni:
For my point of view this deallocation procedure is obvious from ffmpeg documentation
http://ffmpeg.org/doxygen/2.5/avio_8h.html#a853f5149136a27ffba3207d8520172a5
```The buffer must be allocated with av_malloc() and friends. It may be freed and replaced with a new buffer by libavformat. AVIOContext.buffer holds the buffer currently in use```
so we can't rely on `self.buffer` as live memory, `self.iocontext.buffer` is more correct pointer to be freed

heni:
**And more detailed description for the whole \_\_dealloc\_\_ patch**
~~~
diff --git a/av/container/core.pyx b/av/container/core.pyx
index 3f2744d..097d878 100644
--- a/av/container/core.pyx
+++ b/av/container/core.pyx
@ -106,20 +107,22 @@ cdef class ContainerProxy(object):
    def __dealloc__(self):
         with nogil:

             # Let FFmpeg handle it if it fully opened.
~~~
  We can't rely on `self.ptr and not self.writeable` because it's not garantee success result of `lib.avformat_open_input` call single possible way to introduce boolean flag
that we raise immediately after sucessfull `lib.avformat_open_input` call
~~~
-            if self.ptr and not self.writeable:
+            if self.is_opened:
                 lib.avformat_close_input(&self.ptr)
~~~
 `self.buffer` can be reallocated by `iocontext `related method (according with ffmpeg documentation), so we can free it only in case when `iocontext` is not allocated,
otherwise we have to free `iocontext.buffer` (by ffmpeg documentation) and of course `iocontext`
~~~
             # Manually free things.
             else:
-                if self.buffer:
-                    lib.av_freep(&self.buffer)
                 if self.iocontext:
+                    lib.av_freep(&self.iocontext.buffer)
                     lib.av_freep(&self.iocontext)
+                else:
+                    lib.av_freep(&self.buffer)
~~~

mikeboers:
Thank you very much for humoring me. This will go a long way to me finally feeling confident about this.

mikeboers:
Thanks again for the explanation. I've commented it a bit to make myself more comfortable with it, and added another `avformat_free_context` to the end because I think that might be necessary for outputs.

This is now merged.

🎉

PR #258, opened 2017-07-19
Enable reading Unicode metadata

rr-:
https://github.com/mikeboers/PyAV/issues/197

I'm new to Cython so the code is kludgy, but it can now handle both ASCII and UTF-8 metadata. It will bork for non-UTF-8 metadata.


mikeboers:
Thanks for your efforts, but I've just merged another PR which did much the same thing but a bit more robustly -> https://github.com/mikeboers/PyAV/pull/250

Cheers!

:beers:

Mike

PR #263, opened 2017-08-04
Expose frame corruption check.

Litterfeldt:
`int AVFrame::flags` is a collection of `AV_FRAME_FLAGS`

Added property `is_corrupt` to class Frame that checks
if `AV_FRAME_FLAG_CORRUPT` is set which would then indicate that:

```
The frame data may be corrupted, e.g.
due to decoding errors.
```


mikeboers:
Done and done.

Thank you. =]

PR #267, opened 2017-08-18
Fix Container property initialization order

mkfmnn:
metadata_encoding needs to be set before the ContainerProxy is
instantiated, because it's read in the proxy's initializer.

Fixes mikeboers/PyAV#265


mikeboers:
Thanks. Sorry it took me a week. ;)

PR #284, opened 2018-02-13
Warnings fixed

rawler:


mikeboers:
I pulled this one at the same time. Thank you for this, and your patience!

PR #304, opened 2018-04-02
fix Stream options auto init empty dict

axmadjon:


axmadjon:
why shows an failed? what is the problem?

PR #327, opened 2018-05-17
Added information about the frame is it interlaced or progressive.

MPGek:
This is useful when you need to decide to filter interlaced artifacts or not.


mikeboers:
Thanks! Sorry it took me so long, but I just merged this via a cherry-pick.

PR #330, opened 2018-05-27
Fix IndentationError(unexpected indent) in activate.sh

gemfield:
When do `bash scripts/build-deps`,the script will print below errors:
`+++ python -c '
    from __future__ import print_function
    import sys
    import platform
    print("{}{}.{}".format(platform.python_implementation().lower(), *sys.version_info[:2]))
    '
  File "<string>", line 2
    from __future__ import print_function
    ^
IndentationError: unexpected indent`

This PR is to fix this issue.


gemfield:
Please be noted that CI failed has no relation with this PR.

mikeboers:
Great. Thank you.

PR #332, opened 2018-06-04
a more robust way of finding the right codec

adavoudi:
Sometimes the functions `avcodec_find_encoder_by_name` or `avcodec_find_decoder_by_name` can't find the right codec while it exists and the `OutputContainer` class can find it! 
I noticed the way the `OutputContainer` class find the codec and use the same approach for the `Codec` class.

### DIFF
--- a/av/codec/codec.pyx
+++ b/av/codec/codec.pyx
@@ -48,9 +48,23 @@ cdef class Codec(object):
             return
 
         if mode == 'w':
-            self.ptr = lib.avcodec_find_encoder_by_name(name)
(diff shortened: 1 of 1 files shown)


mikeboers:
I'm working on top of this locally, and going to refactor the `OutputContainer` to hijack `Codec`'s implementation.

I'd love to have a test that would fail without this.

Do you have an example of a codec that would not have been found just by `avcodec_find_*coder_by_name`?

mikeboers:
This was cherry-picked and merged.

PR #347, opened 2018-07-14
Add an example of a simple indexable video wrapper.

nlgranger:
This is only the code for the class, you can suggest how to turn this into a script if you have any idea to do so. Other requests to improve it are welcome of course.

The use case is people who want to avoid OpenCV which is quite big and tedious to install but still want to avoid loading complete videos when working on them. I think only PyAV offers that possibility but I find it low-level and difficult to use for this simple frame reading task. This wrapper should make it simpler for those interested.

Please double check the code if you plan to merge it, I am by no means an expert of ffmpeg.

### DIFF
--- a/examples/simple_video_wrapper.py
+++ b/examples/simple_video_wrapper.py
@@ -1,0 +1,94 @@
+from typing import Sequence
+import av
+
+
+class VideoSequence(Sequence):
+    """A video wrapper with index based frame access.
+
+    The current behaviour is to return an array of RGB triplets for each frame.
+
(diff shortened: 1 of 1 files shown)


mikeboers:
I've never really engaged with this, because there are a lot of small things that add up to something that is insurmountable. In principal, I'm not against this, but it makes too many assumptions for me to be comfortable (i.e. that you can calculate the duration, and that seeking behaves at all).

I'd be happier if those assumptions were spelled out, and this was better documented, but I'm sorry to say I'm not going to take the effort myself.

Issue #379, opened 2018-09-20
Many build warnings.


mikeboers:
- [ ] ```src/av/stream.c:2094:29: warning: 'codec' is deprecated [-Wdeprecated-declarations]
  switch (__pyx_v_c_stream->codec->codec_type) {```
- [ ] ```src/av/stream.c:2855:15: warning: assigning to 'struct AVCodec *' from 'const struct AVCodec *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
    __pyx_t_7 = __pyx_v_self->_codec_context->codec;```
- [x] ```src/av/packet.c:2394:9: warning: 'av_free_packet' is deprecated [-Wdeprecated-declarations]
        av_free_packet((&__pyx_v_self->__pyx_struct));```
- [ ] ```src/av/_core.c:1574:3: warning: 'av_register_all' is deprecated [-Wdeprecated-declarations]
  av_register_all();```
- [ ] ```src/av/_core.c:1601:3: warning: 'avcodec_register_all' is deprecated [-Wdeprecated-declarations]
  avcodec_register_all();```
- [ ] ```src/av/_core.c:1610:3: warning: 'avfilter_register_all' is deprecated [-Wdeprecated-declarations]
  avfilter_register_all();```
- [ ] ```src/av/format.c:1634:28: warning: 'av_oformat_next' is deprecated [-Wdeprecated-declarations]
      __pyx_v_self->optr = av_oformat_next(__pyx_v_self->optr);```
- [ ] ```src/av/format.c:3353:32: warning: 'av_iformat_next' is deprecated [-Wdeprecated-declarations]
    __pyx_v_2av_6format_iptr = av_iformat_next(__pyx_v_2av_6format_iptr);```
- [ ] ```src/av/logging.c:3694:20: warning: assigning to 'char *' from 'const char *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
      __pyx_v_name = __pyx_v_cls->item_name(__pyx_v_ptr);```
- [ ] ```src/av/descriptor.c:1578:13: warning: assigning to 'struct AVOption *' from 'const struct AVOption *const' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
  __pyx_t_1 = __pyx_v_self->ptr->option;```
- [ ] ```src/av/video/codeccontext.c:2559:21: warning: 'avcodec_encode_video2' is deprecated [-Wdeprecated-declarations]
[...]

PR #380, opened 2018-09-20
Eliminate some compilation warnings (#379)

jlaine:
I'd like to fix a couple more warnings before merging.

### DIFF
--- a/av/video/format.pyx
+++ b/av/video/format.pyx
@@ -128,7 +128,7 @@ cdef class VideoFormatComponent(object):
     property bits:
         """Number of bits in the component."""
         def __get__(self):
-            return self.ptr.depth_minus1 + 1
+            return self.ptr.depth
 
     property is_alpha:
         """Is this component an alpha channel?"""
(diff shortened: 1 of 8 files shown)


av/video/format.pyx
```diff
@@ -128,7 +128,7 @@ cdef class VideoFormatComponent(object):
     property bits:
         """Number of bits in the component."""
         def __get__(self):
-            return self.ptr.depth_minus1 + 1
+            return self.ptr.depth
```
mikeboers: Ahaha what a change.

jlaine: I'm really good at math huh? ;)

jlaine: Ah I forgot to remove the comment referring to a bitfield, these are just plain ints

jlaine:
Concerning *_register_all the deprecation only starts with ffmpeg 4.0. i need to check whether they were already no-oops for 3.x (in which case we can drop them) or whether we need some conditional logic.

jlaine:
I had closer look at audio_rates / frame_rates, I'll move the discussiion over to a separate PR

jlaine:
If the rest of the changes look good, can get this first round merged?

mikeboers:
Yeah. I'll do it in a few hours. 😁

PR #381, opened 2018-09-21
Fix Codec.audio_rates and Codec.frame_rates

jlaine:
The properties returned a pointer cast to an int instead of the
actual supported rates.

I have added a test using Opus to test audio_rates.

I have not found *any* codecs which report frame_rates,
so that property is untested.

### DIFF
--- a/av/codec/codec.pxd
+++ b/av/codec/codec.pxd
@@ -1,15 +1,13 @@
-from libc.stdint cimport uint64_t
-
 cimport libav as lib
 
 
 cdef class Codec(object):
 
-    cdef lib.AVCodec *ptr
-    cdef lib.AVCodecDescriptor *desc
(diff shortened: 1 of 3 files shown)


mikeboers:
AppVeyor declares that `opus` is actually `libopus`. 😝

jlaine:
It's a mess, some ffmpeg have no opus encoder at all, on Windows libopus masquerades as opus.

I think I've got it right now, let's see what CI says.

jlaine:
Actually given how slow CI is maybe merge #380 and this one and I'll circle back to #382 later.

mikeboers:
I cherry-picked, ammended, and merged them all.

mikeboers:
I'm living dangerously and pushing my out-of-order commits before CI says it is good. 🤠

PR #382, opened 2018-09-21
Rename avrational_to_faction to avrational_to_fraction

jlaine:
There was an "r" missing. Also, make its input a const pointer to avoid
compilation warnings.


mikeboers:
Ammended, and merged.

PR #387, opened 2018-09-22
Reduce const warnings

jlaine:
AVCodec pointers should always be passed as `const AVCodec*`.

This also fixes a few other missing const qualifiers.

The remaining const warnings are trickier as they are mostly a product of how Cython generates its code.


jlaine:
Haha, no problem, I found another couple of easy pickings which I'll put in another PR

PR #395, opened 2018-09-25
Stop messing with packet dts and pts

jlaine:
A packet's timebase is that of the stream it belongs to. Storing the
timebase and manipulating it ourselves leads to headaches.

For instance when encoding video, the "flush" packets carry a valid dts
and pts but their timebase doesn't get set, causing _rebase_time to
destroy these timestamps. This leads to an exception upon flushing the
output stream.

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -334,7 +334,6 @@ cdef class CodecContext(object):
         ):
             for frame in frames:
                 for packet in self._send_frame_and_recv(frame):
-                    self._setup_encoded_packet(packet, frame)
                     res.append(packet)
             return res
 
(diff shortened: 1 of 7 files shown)


mikeboers:
I don't believe your premise is correct. During decoding, `AVPacket.pts` is in `AVCodecContext.time_base`, and during encoding it must be in `AVStream.time_base`.

Further, having a packet fully describe its own time potentially allows for much easier remuxing as a packet can be passed from one stream to another.

You're going to need to convince me of this one.

jlaine:
Yeah I missed the case where we move packets from one stream to another, as it wasn't capture by any tests. I'm closing this and I'll try to narrow down the specific issue I have.

jlaine:
One thing that totally confuses me is the fact that both a *stream* and a *codec context* carry a timebase. Where things go completely awry is how the Stream.time_base accessors are wired up:

- the getter reads from self._stream._timebase

- the setter is *never* invoked, as `__setattr__` swallows it, and stores the result into self.codec_context.timebase

As a result you cannot even round-trip a get/set:

```
>>> import fractions
>>> import av
>>> container = av.open('foo.mp4', 'w')
>>> stream = container.add_stream('h264', rate=30)
>>> stream.time_base
Stream.time_base -> None
>>> stream.time_base = fractions.Fraction(1, 90000)
Stream.__setattr__("time_base")
>>> stream.time_base
Stream.time_base -> None
```

-> stream.time_base is pretty much useless, or at least very dangerous

Issue #397, opened 2018-09-27
Packet "time_base" manipulations are tricky


jlaine:
While trying to encode frames (with a correct PTS)  to an H.264 stream I kept encountering problems when I submit the "flush" packets to the muxer, complaining that the packets had non-monotonically increasing PTS.

After some digging around I have come to the conclusion that PyAV's storing of a "time_base" associated with packets and its attempts to manipulate the pts / dts / duration are harmful.

The root of the problem comes from how PyAV attempts to associate a "time_base" with a packet (a concept that ffmpeg's AVPacket does not have) during encoding. In essence CodecContext.encode works like this:

- send a frame to the encoder
- check if it has some packets for us
- if it does, try to put either the frame's time_base or the stream's into the packet (setup_encoded_packet)
- return the packets

The underlying assumption here is that the packets we received are related to the frame we just submitted, but *this is not necessarily the case*. The reason there is a "flushing" mechanism is precisely because the codec may be performing buffering.

H.264 is a good example here : if you look at the timeline you will see there is clearly some buffering going on. You get no packets for a long time even though you're submitting frames, and conversely the flushing produces packets even though you didn't submit a frame.. so setup_encoded_packet is working on a wrong assumption.

Also it's not clear to me why setup_encoded_packet isn't crashing when given a "None" frame, as we reach into frame._time_base?

jlaine:
My conclusion at this stage are that:

- at the very least setup_encoded_packet should not be looking at "frame" at all. This is doubly true as CodecContext.encode accepts a single frame a list of frames or None, and reassigns the "frame" variable while iterating

- I am not convinced we are helping users at all by mangling pts/dts by default. As far as I can tell the only use case where this makes sense is when transferring packets between streams, which is definitely an advanced topic

jlaine:
Here is a sample piece of code for tests:

```python
output = av.open('some.mov', 'w')
  
stream = output.add_stream('h264', 24)
stream.width = WIDTH
stream.height = HEIGHT
stream.pix_fmt = 'yuv420p'

for i in range(DURATION):
    frame = VideoFrame(WIDTH, HEIGHT, 'rgb24')
    frame.pts = i * 2000
    frame.time_base = Fraction(1, 48000)

    print('frame time_base', frame.pts, 'pts', frame.pts)
    for packet in stream.encode(frame):
        print('stream time_base', stream.time_base)
        print('packet time_base', packet.time_base, 'pts', packet.pts)
        output.mux(packet)

[...]

PR #398, opened 2018-09-27
Fix packet timebase handling during encoding.

jlaine:
We must not use the frame's timebase as frame is None while flushing the
encoder, which gives random results. In any case we know that frames were
rebased into the CodecContext's time_base.

For a lengthier discussion see #397


jlaine:
Problem solved by adding libx264 support to all our Travis builds (appveyor already had it).

As a nice side effect, we no longer skip `test_encoding_h264`.

mikeboers:
I'm really liking this as our immediate solution.

I still want to go on that deep dive to finally decide, once and for all, how time works. But we don't have to do it today.

jlaine:
I agree - let's fix the immediate issue while retaining the overall behaviour and take a breath before we dive deeper. I'm now looking into the subleties of AVCodecContext.time_base vs AVStream.time_base

PR #402, opened 2018-10-01
Expose AVFrame.pict_type

jlaine:
This makes it possible to force an encoder to generate full intra
frames, for instance to recover from picture loss in realtime
communications.

NOTE: I would appreciate guidance as to how the enum values should be exposed

### DIFF
--- a/av/video/frame.pyx
+++ b/av/video/frame.pyx
@@ -254,6 +254,14 @@ cdef class VideoFrame(Frame):
         """Is this frame an interlaced or progressive?"""
         def __get__(self): return self.ptr.interlaced_frame
 
+    @property
+    def pict_type(self):
+        return self.ptr.pict_type
+
+    @pict_type.setter
+    def pict_type(self, value):
+        self.ptr.pict_type = value
+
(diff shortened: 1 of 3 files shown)


jlaine:
I noticed there is an `av_get_picture_type_char`, maybe we should expose this instead of the integer value?

mikeboers:
The ints will play nicely with the enums; see the `pict_type` branch.

I've made the enums able to support multiple names, defaulting to one of them. So you can set `frame.pict_type = 'P'` or `frame.pict_type = 'PREDICTED'` although it will come out being the long version no matter what.

Maybe I should make the instances have multiple names. Hmm.

mikeboers:
Aaaand I've made it so that if you're checking for string equality, it will roundtrip:

```
>>> import av
>>> f = av.VideoFrame()
>>> f.pict_type = 'I'
>>> f.pict_type == 'I'
True
>>> f.pict_type
<PictureType:INTRA(1)>
```

mikeboers:
I've gone and rebased and merged this.

PR #403, opened 2018-10-01
Use av_get_media_type_string

jlaine:
Now that we have dropped libav, we can use this function instead of
defining our own.


mikeboers:
I was thinking of turning this into one of our enums at some point, but is is immediately cleaner.

Hooray for no more Libav!

I spent a little time looking at how FFmpeg versions itself, now that we can trust that, and we could replace the entire reflection process with that. It is wonderful.

PR #405, opened 2018-10-01
Add setters for CodecContext.max_bit_rate and CodecContext.min_bit_rate

jlaine:


jlaine:
I'm not convinced we should merge this, initial testing shows you can get very far with the `options`.

PR #409, opened 2018-10-02
Add more docstrings, hide deprecated members.

jlaine:


mikeboers:
Merged, built, and uploaded.

PR #412, opened 2018-10-04
Fix deprecated codec / format / filter iterators

jlaine:
Only use av_iformat_next and av_oformat_next when using FFmpeg < 4.0,
see:

https://github.com/FFmpeg/FFmpeg/blob/ebc3d04b8df4e11b1343090fed1014832d5cf46d/doc/APIchanges#L92

We can also stop detecting and stubbing of these function which have
been present since at least FFmpeg 2.0:

- avformat_alloc_output_context2
- avformat_close_input


mikeboers:
I'm scared of removing so much of the autoconf-like machinery, but that is the point. LOL

I'm not keen to start implementing stuff (beyond dealing with version differences) in C. I'd much rather expose little `pyav_*` stub functions that do something that mirrors what the FFmpeg C API does, and use those in Cython.

I am keen to hear a counter argument.

If we do start implementing stuff in C, we ought to move those into `*.c` files alongside the Cython, and embed them into specific modules, rather than being in the headers. That isn't enough to stop this PR, because that is new machinery to figure out. I'd be okay to merge this aspect and work on it in the future.

Ideally anyways... I'd like to see all of this in Cython at runtime. That would require that we allow unresolved symbols, which I think is the way we go anyways. Same deal where that isn't enough to stop this PR, as it does go in a good direction, and the new machinery would be a big deal™.

Kamekameha:
I as well believe that is better to implement these in Cython, probably in `*.pxd`-only files as inlined functions and in a different namespace from the `libav` one as they're quite specific to Python/PyAV (I like to see _these_ `*.pxd` files as just like what they are here, Cython-specific include files limited to expose their underlying API). Also, a smart enough compiler should be able to factor out dead constant paths, I think, given these version checks only use macros.

But beside all that, an actual issue I'm seeing is that none of the `pyav_get_available_{codecs,filters,formats}` functions are propagating exceptions at all, and each of the Python/C API functions used in these three functions can indeed raise exceptions; this could segfault easily I presume.

<sub>(Lastly, and not as important, isn't symbol visibility public by default on gcc?)</sub>

jlaine:
How do you #ifdef code in Cython @Kamekameha so that *it is not even compiled* ? If you have a solution which can avoid compiling at all against the deprecated methods / structs, I'm all ears.

@mikeboers feature detection *at runtime* is a completely different kettle of fish. This would require:

- not linking against FFmpeg libraries *at all*
- `dlopen`-ing the libraries at runtime, and looking up all symbols manually

Having worked on Qt's network code that does precisely this to avoid linking against a specific OpenSSL version I can tell you it is gigantic amount of work and very prone to breakage.

Kamekameha:
You can't, I didn't imply that. That's why I said that last statement in my first paragraph. Do note that I'm not against implementing stuff directly in C, but that it should work out quite alright if done solely in Cython and that I lean more towards that.

mikeboers:
Gonna try to weigh on on a pile of stuff at the same time.

> I agree we need to keep these bits of C absolutely minimal - and as you can see the bits of code are trivially short.

I don't think they are trivial as soon as they start hitting the Python API. I'm personally not comfortable there, and to be fair @Kamekameha pointed out that they aren't interacting with exceptions properly (I'm taking that as given without any authority on my part).

> I am absolutely positive the maintenance of these bits of C is far lower than the headaches posed by an "introspection" mechanism, redefining function prototypes in .pxd files and stubbing out ffmpeg's function (which was done in C by the way!).

I agree with the first part for sure. I think that a `pyav_demuxer_iterate` which picks between `av_demuxer_iterate` and `av_iformat_next` is great and fine. This is what your `pyav_find_output_format` does. I don't like the part of `pyav_get_available_formats` that deals with sets. My whole deal here with this is that I would rather see that part in Cython.

> I as well believe that is better to implement these in Cython, probably in *.pxd-only files as inlined functions and in a different namespace from the libav one as they're quite specific to Python/PyAV

I don't get this. FFmpeg's are `av*` and ours are `pyav*`. They both get c-imported as the same `lib` in Cython, but that is something we could change without significant effort or effect.

> Also, a smart enough compiler should be able to factor out dead constant paths, I think, given these version checks only use macros.

That would be nice, but isn't what I've seen.

> It's more of a project's structure thing. If anything, *.pxd files (both from the libav and av packages) could be installed alongside the binaries in the future if their Cython "API" gets ever exposed for other Cython projects. Also, it makes the most sense to me to put clearly PyAV (or Python for that matter) specific stuff in the av directory at least.

I'm not sure about the specifics of how you intend this to work, but I would also like to see these "shims" (and whatever else they're called) be in either `*.c` files (or something Cython), and integrated into specific Python modules, rather than be functions defined in headers pulled into everything. I think I said this at the top. But I also said it isn't enough to stop me from merging this, opening another issue discussing what we would want to do, and dealing with it later as I think this falls into the category of bike-shedding.

mikeboers:
Gonna switch to http://docs.cython.org/en/latest/src/userguide/external_C_code.html#implementing-functions-in-c ; sit tight!

mikeboers:
This is much nicer feeling that my earlier version! https://github.com/mikeboers/PyAV/commit/0f0dc124571339d68ebe5c71e5a96596a211cba8

jlaine:
If you want to move the code to a .c file that's fine by me. However bear in mind there is a single call site for each of these functions, so inlining doesn't incur any cost.

Kamekameha:
> They both get c-imported as the same `lib` in Cython, but that is something we could change without significant effort or effect.

Yeah that, what I meant is to have them separated, pretty much what you just did in your proposal commits.

> I'm not sure about the specifics of how you intend this to work

I guess this follows through with what your shims do. My idea is having support for something along the lines of what http://docs.cython.org/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules suggests (at the end of the section).

> as I think this falls into the category of bike-shedding.

Agree, really. The only real concern I have specific to this PR is what I mentioned about exceptions, everything else is just my opinion.

mikeboers:
@jlaine I'm not a very C person. Don't they need to be `static` for that or something?

I dunno folks. There are a few issues in here:

1. Using the Python API in C (not addressed in my test above, and yet this is the only one I feel strongly about).
2. Deciding that these are more PyAV than adaptors to FFmpeg, and to move all the PyAV code together.
3. How to actually implement it.

Sounds like there isn't even any strong opinions about 2/3 at this time. If someone wants to pick one and/or flip a coin, I'm game. I'm too tired to made arbitrary decisions today.

jlaine:
I'm going to let you call it - to be honest I was quite happy with the code as I submitted it : it fixed the deprecation warnings, works, and eliminated 100 lines of code ;)

mikeboers:
I went ahead and moved what I wanted into C files beside the modules, and did the Python manipulations in Cython.

_shrugs_

Waiting for CI then it will be pushed to `develop`.

mikeboers:
Merged, after several passes of rewriting history to fix my own shit.

I am very not used to syntax errors due to macro checks. I'm going to actually start checking locally against FFmpeg 3.3 and 4.0 when working on this stuff, because that was silly of me.

jlaine:
I can't say I'm a fan of what you've done here at all, pretending that `av_demuxer_iterate` and `av_iformat_next` have the same semantics. If you look at how FFmpeg implements its backwards compatibility, it's significantly more involved:

https://github.com/FFmpeg/FFmpeg/blob/1e20ed4382b0d99226d58bea82ace14fd3286e4e/libavformat/allformats.c#L574

`pyav_muxer_iterate` is unsafe to use in a general context, so I'm still of the opinion we should have encapsulated the whole "get all available formats / codecs / filters" in a C function. Failing that I'd rather just go back to the deprecated function calls.

PR #413, opened 2018-10-04
Always use send / recv mechanism

jlaine:
Since FFmpeg 3.1, the preferred mechanism for encoding / decoding audio and video is using a send / receive mechanisms instead of calling avcodec_xxx_encode and avcodec_xxx_decode methods.

Now that our baseline is FFmpeg 3.2 we can remove the legacy code, which fixes a number of deprecation warnings.

### DIFF
--- a/av/audio/codeccontext.pyx
+++ b/av/audio/codeccontext.pyx
@@ -58,23 +58,6 @@ cdef class AudioCodecContext(CodecContext):
     cdef Frame _alloc_next_frame(self):
         return alloc_audio_frame()
 
-    cdef _decode(self, lib.AVPacket *packet, int *data_consumed):
-
-        if not self.next_frame:
-            self.next_frame = alloc_audio_frame()
-
-        cdef int completed_frame = 0
-        data_consumed[0] = err_check(lib.avcodec_decode_audio4(self.ptr, self.next_frame.ptr, &completed_frame, packet))
-        if not completed_frame:
-            return
-
-        cdef AudioFrame frame = self.next_frame
-        self.next_frame = None
-
-        frame._init_user_attributes()
-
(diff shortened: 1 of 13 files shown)


jlaine:
One thing I realised is that the `count` argument for decoding was completely ignored when decoding with send / recv, so I removed it. This makes me think that `Packet.decode_one()` is a bad idea and should die.

mikeboers:
Now that we have removed the legacy methods, I'm rethinking the whole setup, and that the PyAV API should mirror the FFmpeg API.

I'm happy to leave a `Packet.decode()` method at all, but I think exposing a `send/recv` on the `CodecContext` (and maybe `Stream`) would be a better alternative to the `count` parameter.

That aside, I'm torn on dropping the `count`. The one circumstance in which I know it was there for was to force PyAV to stop polling because FFmpeg would block when polling for Frames. There was only one frame to remove, and the user knew that, so nothing was lost.

Personally, I'd leave it (and/or re-implement it) until we decide to have something mirroring the `send`/`recv`, and implement that.

jlaine:
I just want to point out that this commit doesn't change the behaviour compared to 0.5.3 : we were already using send / recv, so the "count" parameter did nothing.

So the immediate question is : do we keep the "count" parameter as a no-op or be honest and drop it?

Concerning the "blocking" behaviour, I don't think it's justified, if you need your main loop to be responsive you should be running your `demux()` or `decode()` call in a thread. The blocking code occurs within a C function call and the GIL is released. I for instance use this in aiortc's "webcam" demo.

mikeboers:
Merged. Just waiting for CI to pass before pushing to `develop`.

mikeboers:
Merged.

PR #414, opened 2018-10-04
Deterministically perform bytes -> str decoding

jlaine:
On Python 2, the previous code did not systematically perform the
decoding depending on whether or not an encoding was given.

With this PR we align the Python 2 behaviour on the sane behaviour of
Python 3. This also avoids a compilation warning on Python 3 were we
tested "str is str".

### DIFF
--- a/av/utils.pyx
+++ b/av/utils.pyx
@@ -113,13 +113,8 @@ cpdef int err_check(int res=0, filename=None) except -1:
 # === DICTIONARIES ===
 # ====================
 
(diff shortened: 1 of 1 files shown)


mikeboers:
Waiting on CI.

mikeboers:
Merged.

(Do you get notifications of the merge without me saying "Merged."?)

jlaine:
Yes I do, you can save some typing!

Issue #77, opened 2015-01-30
How to make VideoFrame.to_nd_array() return an ndarray of the correct shape and dtype?


ksze:
Again, working with my FFV1 16-bit grayscale videos, which have, of course, only one grayscale channel.

If I do this:

```
In [25]: container = av.open('/home/ksze/16-bit grayscale ffv1.avi')

In [26]: _break = False

In [27]: for p in container.demux():
    for f in p.decode():
        if isinstance(f, av.VideoFrame):
            frame = f
            _break = True
            break
    if _break:
        break
   ....:

In [28]: frame.to_nd_array()
Out[28]:
array([[[ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0],
        ...,
        [ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0]],

       [[ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0],
        ...,
        [ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0]],

       [[ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0],
        ...,
        [ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0]],

       ...,
       [[ 0,  0,  0],
        [16, 16, 16],
        [ 0,  0,  0],
        ...,
        [ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0]],

       [[ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0],
        ...,
        [ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0]],

       [[ 0,  0,  0],
        [ 0,  0,  0],
        [ 0,  0,  0],
        ...,
        [ 0,  0,  0],
        [11, 11, 11],
        [ 0,  0,  0]]], dtype=uint8)
```

Note that I now that an ndarray that represents _three_ channels and with the wrong dtype.

And then I tried to do this:

```
In [29]: frame.to_nd_array(colorspace='gray16le')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-29-4d71a13e89d1> in <module>()
----> 1 frame.to_nd_array(colorspace='gray16le')

/home/ksze/PyAV/av/video/frame.so in av.video.frame.VideoFrame.to_nd_array (src/av/video/frame.c:3515)()

[...]

PR #415, opened 2018-10-06
Add support for more ndarray <-> VideoFrame conversions

jlaine:
We support round-tripping:

- 24-bit RGB
- 32-bit RGBA (new)
- 8-bit grayscale (new)
- yuv420p (new)
- yuyv422 (new)

We drop the gray16 conversion code as it was not handling byte order
correctly, and could not be triggered anyway due to an impossible test:

  frame.format.name == ('gray16le', 'gray16be')

### DIFF
--- a/av/video/frame.pxd
+++ b/av/video/frame.pxd
@@ -17,7 +17,7 @@ cdef class VideoFrame(Frame):
 
     cdef readonly VideoFormat format
 
-    cdef _init(self, lib.AVPixelFormat format, unsigned int width, unsigned int height)
+    cdef _init(self, lib.AVPixelFormat format, unsigned int width, unsigned int height, unsigned int align)
     cdef _init_user_attributes(self)
 
     cdef _reformat(self, unsigned int width, unsigned int height, lib.AVPixelFormat format, int src_colorspace, int dst_colorspace)
--- a/av/video/frame.pyx
+++ b/av/video/frame.pyx
@@ -32,6 +32,21 @@ cdef EnumType PictureType = define_enum('PictureType', (
 ))
 
 
+cdef useful_array(VideoPlane plane, bytes_per_pixel=1):
+    """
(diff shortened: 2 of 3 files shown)


jlaine:
I found #77 which seems to be where the 16-bpp grayscale code originated, but unfortunately the link to the sample file seems broken. Does the FATE suite contain any such examples?

mikeboers:
Cherry-picked.

mtn:
I'm not sure I understand this issue fully -- should these frames be converted to and from yuv420p without changing the underlying data?

```python
>>> import numpy as np
>>> from av import VideoFrame
>>> a = np.zeros((3,3,3))
>>> a[:,:,2] = 255
>>> a = a.astype(np.uint8)
>>> a
array([[[  0,   0, 255],
        [  0,   0, 255],
        [  0,   0, 255]],

       [[  0,   0, 255],
        [  0,   0, 255],
        [  0,   0, 255]],

       [[  0,   0, 255],
        [  0,   0, 255],
        [  0,   0, 255]]], dtype=uint8)
>>> video_frame = VideoFrame.from_ndarray(a, format="bgr24")
>>> video_frame.reformat(format="yuv420p").reformat(format="bgr24").to_ndarray()
array([[[  0,   0, 253],
        [  0, 230,   0],
        [  0, 136,   0]],

       [[  0,   0, 253],
        [  0, 230,   0],
        [  0, 136,   0]],

       [[  0,   0, 253],
        [  0, 230,   0],
        [  0, 136,   0]]], dtype=uint8)
>>>
```

jlaine:
Roundtrip conversion between RGB and YUV is unfortunately bound to introduce some level of error (plane decimation, rounding / truncation).

I think in your example the decimation is making the biggest difference, especially as your image's dimension (3x3) are both small and odd. Slightly changing your example as follows already gives less error on the roundtrip:

`a = np.zeros((4,4,3))`

Issue #416, opened 2018-10-07
Strip the `reflect` setup.py command.


mikeboers:
We no longer have a use for it, and I suspect that the build-time version checking will be done another way.

We could/should strip out that functionality, and everything in `setup.py` that existed to support it.

I think that we should, for a while, leave a `reflect` command that does nothing except indicate that it is deprecated. I think this because I've left quite a few messages around about running the reflect command, and I'd like it not to fail hard on people.

jlaine:
With reflect out of the way, we may actually be able to support builds against a static ffmpeg - which would be my preferred option for distributing a binary wheel.

mikeboers:
That would be wacky!

My past experience with that sort of things is with [PyTomCrypt](https://github.com/mikeboers/pytomcrypt) (which was my first Cython-wrapping project). It compiles LibTomCrypt itself, and links it into the `_core` module. At first import, it [uses ctypes to assert that symbols are resolved globally](https://github.com/mikeboers/PyTomCrypt/blob/master/tomcrypt/__init__.py#L14-L20) (or something something).

PR #418, opened 2018-10-08
Eliminate "reflect" command (fixes: #416)

jlaine:
We no longer perform any kind of "reflection", we simply use the FFmpeg
library versions at compile time.


jlaine:
Let's tread carefully on this one, I'd appreciate your insights on how the library / include paths are set in the non-pkg-config case.

mikeboers:
I don't have insights on how they work without `pkg-config`. I think the idea is that it just assumes it has everything it needs via `CFLAGS`, etc..

jlaine:
OK, in that case I think our current setup is sane. The setup.py is a lot easier to read..

PR #419, opened 2018-10-08
Fix constness warnings for AVOption

jlaine:
All AVOption manipulation should be done on const pointers, including
the AVClass.option definition, see:

https://www.ffmpeg.org/doxygen/4.0/structAVClass.html


mikeboers:
Cherry-picked.

PR #420, opened 2018-10-08
Prune some unreachable code

jlaine:
The AVPacket.copy() method did nothing useful.


av/utils.pyx
```diff
@@ -106,8 +106,6 @@ cpdef int err_check(int res=0, filename=None) except -1:
     else:
         raise AVError(-res, message, None,     log)
 
-    return res
```
jlaine: This is unreachable as both branches of the if / else above raise an error

mikeboers:
Cherry-picked.

PR #421, opened 2018-10-09
Fix buffer_size calculations for AudioFrame planes

jlaine:
This PR fixes several misunderstandings concerning AudioFrame buffer
sizes:

- for audio, "line size" and "buffer size" are equivalent, as there is
only ever a single "line".

- only linesizes[0] is ever populated, and it gives you the buffer size
of each plane.

- we do not need to recalculate anything upon putting frames into an
AudioFrame, the buffer size does not change.

- "align" is not a boolean, it can be any power of two.

We stop carrying around "align", "nb_channels" and "nb_planes", they are
not needed.

### DIFF
--- a/av/audio/frame.pxd
+++ b/av/audio/frame.pxd
@@ -13,10 +13,6 @@ cdef class AudioFrame(Frame):
     cdef uint8_t *_buffer
     cdef size_t _buffer_size
 
-    cdef bint align
-    cdef int nb_channels
-    cdef int nb_planes
-
     cdef readonly AudioLayout layout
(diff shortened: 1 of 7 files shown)


jlaine:
You'll notice that `line_size` is gone from `AudioPlane` as I think it's misleading. I doubt anyone used it since it gave a bogus value of 0 for anything but the first plane. If you absolutely want it back I'd go with a `renamed_attr` pointing to `buffer_size`.

mikeboers:
Cherry-picked.

PR #422, opened 2018-10-09
Add more AudioFrame <-> ndarray conversions

jlaine:


jlaine:
To be discussed : do we keep the previous convention of shape (planes, samples) or do we swap it?

PR #426, opened 2018-10-11
Remove legacy-style buffer protocol from Buffer

jlaine:


jlaine:
Well, that failed spectacularly..

mikeboers:
If I understand the situation:

- Numpy on Py2 uses the old-style buffer.
- The old-style buffer does not have dimensions, although we could hack on strides.
- You said something on Gitter about `the cython code I wrote which copies partial rows in strides is probably just as efficient`, but I don't know what you're referring to. I'm not seeing the copying. Maybe I need more coffee.

jlaine:
I was referring to :

https://github.com/mikeboers/PyAV/blob/304b6f9fcb24156a7ef7c7445dd573fb3f74e949/av/video/frame.pyx#L318

and

https://github.com/mikeboers/PyAV/blob/304b6f9fcb24156a7ef7c7445dd573fb3f74e949/av/video/frame.pyx#L394

mikeboers:
Oh okay cool. I'll admit I've never used the Cython memory view (or whatever) syntax before, and just didn't spot it at a glance. Thanks for dealing with me.

I agree with that being reasonably efficient. The lesson is that PIL/Pillow is not efficient. 😉

jlaine:
I'm closing this PR, we can revisit this at a future date.

Issue #428, opened 2018-10-12
Remove `Stream.seek()` ?


mikeboers:
The semantics are strange to seek on one stream, as it actually modifies all of them.

It could be good to fully remove the feature, as it is confusing.

jlaine:
I now realise we cannot simply pull this API without users losing functionality. Since FFmpeg is the one doing the time base magic, we need some way to pass the stream. Maybe a "stream" argument to Container.seek?

PR #430, opened 2018-10-12
Remove Stream.seek method (fixes: #428)

jlaine:
This method is semantically incorrect, as it makes the whole container
seek to a different position.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,10 @@ Major:
 
 - Dropped support for FFmpeg < 3.2.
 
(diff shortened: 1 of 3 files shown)


jlaine:
Hum. Actually with this we do lose the ability to pass a specific stream index to `av_seek_frame`..

mikeboers:
Maybe a `stream` argument? I don't know. The whole thing is awkward.

jlaine:
I'm going to abandon this PR until the issue is better understood.

PR #436, opened 2018-10-13
Release/6.0.0

mikeboers:
Last pass at the docs before release. Anything else to include?


jlaine:
I made a typo when documenting the AudioFrame changes, the added method is from_ndarray.

Duly noted for the no-manual linebreaks. Can we move these instructions to a "contributing" file? I'm not sure this is the first thing PyAV users should see when reading the changelog

jlaine:
Ah also we should drop or deprecate decode_one as it loses data..

jlaine:
.. and mark as "Stable" in setup.py. PyAV has been production grade for a long time it seems, and the previously "shy" numbering scheme (0.x.y) + Beta status almost drove me away!

mikeboers:
Re. the developer notes: they don't render into HTML; see <http://docs.mikeboers.com/pyav/develop/changelog.html>. Does that make a difference for you?

Re. `decode_one`: I think deprecate until we expose `send` and `recv` themselves.

Re. "Stable"; [I did](https://github.com/mikeboers/PyAV/pull/436/commits/99a8837a3fccaf3732585fb6403c4bffac979442#diff-2eeaed663bd0d25b7e608891384b7298R523).

jlaine:
Concerning the CHANGELOG it's not a show-stopper for me, I just think it would be nice:

- to centralize instructions for new contributors in a single place
- not to have these contributor-related docs show up when you look at the CHANGELOG on github

I quite like cryptography's guide:

https://cryptography.io/en/latest/development/submitting-patches/

setup.py
```diff
@@ -520,18 +520,20 @@ def run(self):
...
        'Operating System :: MacOS :: MacOS X',
        'Operating System :: POSIX',
        'Operating System :: Unix',
+       'Operating System :: Microsoft :: Windows',
        'Programming Language :: Cython',
        'Programming Language :: Python :: 2.7',
        'Programming Language :: Python :: 3.4',
        'Programming Language :: Python :: 3.5',
        'Programming Language :: Python :: 3.6',
+       'Programming Language :: Python :: 3.7',
```
jlaine: Yes! I'd overlooked that..

mikeboers:
I agree with the changelog, and the example you provided. Not going to let that stop us from releasing. Same with docs and examples in general. I'd like to do it, but moving on.

I've added `deprecation.method` to develop, and used it on `Packet.decode_one`.

This branch is now just the version number change.

Good to go (once CI passes)?

Issue #401, opened 2018-09-30
Assert compatible FFmpeg


mikeboers:
Check that we support the given FFmpeg during build, and runtime.

jlaine:
At build time my preferred option would be something along the lines of doing an "import av" and checking it doesn't bomb..

mikeboers:
Lol.

One idea I had was to split up the `build_ext` command into `build_ext_version` and `build_ext_main`, which are exactly the same except that one of them builds a `av._version` module, and the other builds the rest.

The `_version` module burns in the compile-time versions (via macros), and has the runtime versions (via functions).

We can then import it before building the rest of them, check the versions are reasonable, complain if they aren't, and then go on building the rest of them.

jlaine:
Keep it simple, we should not be re-inventing the compiler / linker / dynamic linker's jobs.

mikeboers:
I just want something that is verbose in the case that the user's FFmpeg is too old, as it will save us a lot of GitHub issues.

jlaine:
In that case I'd vote for an #if on LIBAVCODEC_VERSION for example which spews an explicit #error

jlaine:
Haha, I just saw your comment on #393 suggesting exactly the #if / #error combo, nice :)

mikeboers:
You can grab it via `pkg-config`?! Madness.

Will you let me do both? lol

jlaine:
I'd much rather have a single source of truth (the LIB*_VERSION_INT), much easier to maintain.

mikeboers:
_grumble grumble_ You're right.

PR #452, opened 2018-11-06
Assert PyAV is built against a compatible FFmpeg (fixes: #401)

jlaine:
We explicitly fail the build as early as possible if a build is
attempted against an unsupported FFmpeg version. This gives users a
clear indication of what is wrong instead of triggering obscure
compiler errors about missing functions or struct members.

We also remove the reference to PYAV_HAVE_FFMPEG which is not defined
anymore.


jlaine:
@mikeboers are you happy with this?

Issue #299, opened 2018-03-23
TypeError: seek() got an unexpected keyword argument 'mode'


pi-bjl:
Hello,

I see 0.4.0 has just been released. 

Looking at https://github.com/mikeboers/PyAV/compare/v0.3.3...v0.4.0 , It looks like the Stream.seek keyword argument 'mode' has been renamed to 'whence'? 

Also, I don't know if this intentional, but in av/data/stream.pyx it has class DataStream(Stream) with seek method which still has a 'mode' argument.

I guess this should be included in #297, or possibly changed back to 'mode' to keep backwards compatibility.

Thanks!

PR #455, opened 2018-11-07
Remove DataStream.seek override (fixes: #299)

jlaine:
Its signature does not match Stream.seek, leading to the bug mentioned
in #299.

Furthermore, this method does nothing, where as the base Stream.seek
should work for any kind of stream.


mikeboers:
👍

PR #459, opened 2018-11-07
Add a note that PIL/Pillow and Numpy are required

### LINKED ISSUE #91: Warn if there is no PIL/Pillow
From #90, people expect to get PIL/Pillow as a dependency, so we should at least warn at the end of the build output if it is not there.

mikeboers:
Fixes #91.

Simple enough, yeah?


jlaine:
Nit picking : we need the same in `AudioFrame.to_ndarray`

mikeboers:
Agreed. I'll amend the commit.

PR #460, opened 2018-11-07
Testdata

mikeboers:
This is the set of commits that I was talking about that I can't name.

I feel the choice is between `av.testdata` and `av.exampledata` as I don't want to have "sample" in the name so it has zero chance of being confused.

This will be used in the tests (as it already is), and in the examples in the docs (there is one example of it being used in this PR).


jlaine:
Is this still current?

PR #464, opened 2018-11-08
Container.seek(time, stream=foo) for specifying time_base of a stream.

mikeboers:
Pretty straight forward.


jlaine:
Can we add an explicit test for the new method, and make sure the test for the deprecated method issues a warning?

mikeboers:
Done. Cool?

jlaine:
Very cool, that way when we drop the method we already have our good test in place :)

PR #468, opened 2018-11-26
fixed an error in accessing extradata size

adavoudi:
'av.bytesource.ByteSource' object has no attribute 'size'. We need to change it to length in this line.

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -97,13 +97,11 @@ cdef class CodecContext(object):
             else:
                 return None
         def __set__(self, data):
-            self.extradata_source = bytesource(data)
-            free(self.ptr.extradata)
-            buffer_size = self.extradata_source.length 
+            buffer_size = len(data)
+            cdef ByteSource source = bytesource(data)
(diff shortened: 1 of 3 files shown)


jlaine:
Looks like a good catch. Could you please include a minimal unit test (in `tests/test_codec_context.py`) in your PR so this breakage doesn't silently creep back in at a later date?

adavoudi:
Could you please accept this PR without the unit test part as I'm not sure how to create a unit test for this functionality.

jlaine:
If you have reported this issue, you have certainly encountered it, so you are in a unique position to capture the conditions that triggered the problem.

There is no rush to just merge an incomplete PR, let's get it done right.

jlaine:
Please refrain from adding unrelated changes to this PR: keep the PR focused on the initial issue, add tests to it and ensure continuous integration passes so we can move forward.

If this PR turns into a dumping ground for unrelated changes it has no chance of landing.

jlaine:
ping @adavoudi ? Can I just have a simple example of how one uses extradata?

mikeboers:
This was just fixed.

PR #472, opened 2018-11-29
Codec parsing

mikeboers:
Another option is to create a separate `CodecParser` object. While that feels more flexible, etc., when would someone really want to have the parser in isolation from the context?

I'll answer my own question: when someone doesn't care to decode at the moment.

I'm going to go isolate it.


mikeboers:
Nevermind. The thing needs a `AVCodecContext`. I'm leaving it where it is.

What to folks think?

PR #474, opened 2018-12-05
Expose stream level and flags

magicbear:
Expose stream level and flags for remuxing stream.

### DIFF
--- a/av/packet.pyx
+++ b/av/packet.pyx
@@ -204,6 +204,9 @@ cdef class Packet(Buffer):
             if self.struct.duration != lib.AV_NOPTS_VALUE:
                 return self.struct.duration
 
+    property flags:
+        def __get__(self): return self.struct.flags
+
     property is_keyframe:
         def __get__(self): return bool(self.struct.flags & lib.AV_PKT_FLAG_KEY)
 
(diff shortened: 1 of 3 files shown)


av/packet.pyx
```diff
@@ -204,6 +204,9 @@ cdef class Packet(Buffer):
             if self.struct.duration != lib.AV_NOPTS_VALUE:
                 return self.struct.duration
 
+    property flags:
```
jlaine: For new code, we prefer the modern form:

```
@property
def flags(self):
    return self.struct.flags
```

jlaine: Also, @mikeboers has some infrastructure in place for enums, would it be appropriate to use these?

My concern is that simply exposing "flags" as-is means that users will have to hard-code the AV_PKT_FLAG_xx in their code to interpret "flags".

magicbear: I use the flags for remux the stream, just copy it, so using the original flags value

jlaine:
The CI build failure is not due to your changes, there is an issue with OS X builds

WyattBlue:
Closing this because of its age and lack of interest

Issue #319, opened 2018-05-04
incorrect frame.time due to use of codec_time_base


edhalter:
System: Linux, ffmpeg 3.3.6, PyAV from git (last commit on April 10)

Was using PyAV to iterate through a 25 FPS video and found frame1's .pts is 0 and .time is 0.0. All good. But then frame2's .pts is 40 and .time is 1.6 (instead of 0.04). Hm...

av/frame.pyx:
```python
    property time:
        def __get__(self):
            if self.ptr.pts == lib.AV_NOPTS_VALUE:
                return None
            else:
                return float(self.ptr.pts) * self._time_base.num / self._time_base.den
```

That looks fine, but Frame._time_base is evidently set to 1/25, not the 1/1000 it should be (in this case). Using ffprobe terminology, it's calculating time using the stream's codec_time_base instead of the stream's time_base.

Was curious, so then I looked at all the time-related values from ffprobe and tried to match them up w/ values from PyAV. Run these cmds:

```
ffprobe -select_streams v -show_format -show_streams /path/to/video.mp4 2>/dev/null | grep -e '^\[' -e 'duration' -e '_time' -e 'frame_rate' -e 'start' -e 'base' --color=never
ffprobe -select_streams v -show_frames /path/to/video.mp4 2>/dev/null | sed -n '26,50p;51q' | grep -e '^\[' -e 'pts' -e 'dts' -e 'time' -e 'duration' --color=never
```

(Note that the 2nd cmd grabs the 2nd frame's values, since the 1st frame has mostly 0s.) Output is:

```
[STREAM]
codec_time_base=1/25
r_frame_rate=25/1
avg_frame_rate=25/1
time_base=1/1000
start_pts=0
start_time=0.000000
duration_ts=N/A
duration=N/A
[/STREAM]
[FORMAT]
start_time=0.000000
duration=90.600000
[/FORMAT]
```
```
[FRAME]
pkt_pts=40
pkt_pts_time=0.040000
pkt_dts=40
pkt_dts_time=0.040000
best_effort_timestamp=40
best_effort_timestamp_time=0.040000
pkt_duration=N/A
pkt_duration_time=N/A
[/FRAME]
```

Then, in a python console:

```python
import av
path = '/path/to/video.mp4'
container = av.open(path)
stream = container.streams.video[0]
packet = next(container.demux(stream))
frames = packet.decode()
frame = frames[0]
properties = {}
[...]

PR #482, opened 2018-12-26
Add support r_rate from ffmpeg r_frame_rate

KengoSawa2:
Hi Thank you great library!!

I am creating a tool to excel metadata with PyAV.

Current PyAV can refer to avg_frame_rate.
However, when displayed with avg_frame_rate, it shows a value different from popular players.
What I confirmed are Mac's QT player, VLC, djv, mediainfo.

Another problem.
When trying to display "TC" by passing avg_frame_rate to the TimeCode library, it tends to be a strange value and it is difficult to use.

For the above two reasons, I think it would be more convenient to make it possible to access r_frame_rate with PyAV.

#319 may be somewhat related(reason is different from me)

### DIFF
--- a/av/stream.pyx
+++ b/av/stream.pyx
@@ -211,6 +211,15 @@ cdef class Stream(object):
         def __get__(self):
             return avrational_to_fraction(&self._stream.avg_frame_rate)
 
+    property r_rate:
+        """
+        The Real base framerate of this stream.
+
+        :type: fractions.Fraction
+        """
+        def __get__(self):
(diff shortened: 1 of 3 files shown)


mikeboers:
I'm considering renaming this attribute as far as Python is concerned.

`avg_frame_rate` seems reasonable as-is.

Perhaps this should be `base_frame_rate` and `av_guess_frame_rate(...)` should be exposed as `best_frame_rate` or `guessed_frame_rate`

mikeboers:
Gonna keep this going in #564

PR #484, opened 2019-01-25
Add flush method to OutputContainer to force a flush of the buffer

koenvo:
In some cases it's useful to force a flush of the current output buffer. The code is borrowed from the hlsenc.c

### DIFF
--- a/av/container/output.pyx
+++ b/av/container/output.pyx
@@ -191,6 +191,10 @@ cdef class OutputContainer(Container):
 
         self._done = True
 
+    def flush(self):
+        lib.av_write_frame(self.ptr, NULL)
+        lib.avio_flush(self.ptr.pb)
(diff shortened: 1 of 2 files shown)


jlaine:
Could you please elaborate on the "some cases"? Do you want to do this when closing the file? At other times?

Also it looks as though this only applies to the case where Python I/O is used, won't it crash if `proxy.ptr.pb` is not set?

koenvo:
At other times. I want to use pyav to remux a live stream from one container format to another one (fMP4), and pass the output in real time to a downstream application. I want to package the output data per GOP.

Altho I need to use it with the Python I/O it also works when using av.open("filename.ext", "w").

WyattBlue:
I don't really see a reason to have this

PR #496, opened 2019-03-26
Remove ContainerProxy.

mikeboers:
This was a layer that was once required for garbage collection due to the
cyclic relationships between containers and streams. Newer Cythons seem to deal
with it better.

### DIFF
--- a/av/container/core.pxd
+++ b/av/container/core.pxd
@@ -5,24 +5,18 @@ from av.dictionary cimport _Dictionary
 from av.format cimport ContainerFormat
 from av.stream cimport Stream
 
-# Since there are multiple objects that need to refer to a valid context, we
-# need this intermediate proxy object so that there aren't any reference cycles
-# and the pointer can be freed when everything that depends upon it is deleted.
-cdef class ContainerProxy(object):
 
-    cdef bint writeable
+cdef class Container(object):
+
+    cdef readonly bint writeable
     cdef lib.AVFormatContext *ptr
 
-    cdef seek(self, int stream_index, offset, str whence, bint backward, bint any_frame)
-    cdef flush_buffers(self)
-
-    # Copies from Container.
-    cdef object name
-    cdef str metadata_encoding
-    cdef str metadata_errors
+    cdef readonly object name
+    cdef readonly str metadata_encoding
+    cdef readonly str metadata_errors
 
     # File-like source.
-    cdef object file
+    cdef readonly object file
     cdef object fread
     cdef object fwrite
     cdef object fseek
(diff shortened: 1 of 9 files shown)


mikeboers:
Seems like CI is having a hard time with this. But not because the tests are failing, but not finishing.

mikeboers:
So I think the tests are now passing. Unsure if it is because the various CI was having trouble in general, or if this change causes non-deterministic-ish lockups.

jlaine:
AppVeyor still looks pretty unhappy!

jlaine:
Ah. Looks like this branch needs to be rebased following the flake8 stuff, sorry!

av/container/core.pxd
```diff
@@ -36,27 +30,15 @@ cdef class ContainerProxy(object):
...
-
     cdef readonly ContainerFormat format
 
     cdef readonly dict options
     cdef readonly dict container_options
     cdef readonly list stream_options
 
-    cdef ContainerProxy proxy
-    cdef object __weakref__
+    # cdef object __weakref__
```
jlaine: I think this line can go :)

mikeboers:
Yeah... they all do. That was my biggest holdup. But I figure if I flake8 them first, it won't be too bad.

av/stream.pyx
```diff
@@ -87,10 +86,6 @@ cdef class Stream(object):
         self.codec_context = wrap_codec_context(self._codec_context, self._codec, False)
         self.codec_context.stream_index = stream.index
 
-    def __dealloc__(self):
-        if self._codec_options:
```
jlaine: Nice catch, it looks like this was not used at all!

av/stream.pxd
```diff
@@ -3,24 +3,22 @@ from libc.stdint cimport int64_t
...
 from av.frame cimport Frame
 from av.packet cimport Packet
 
 
 cdef class Stream(object):
 
     # Stream attributes.
-    cdef ContainerProxy _container
-    cdef _weak_container
+    cdef readonly Container container
```
jlaine: I'm thinking we need some kind of mechanism to break the Container <-> Stream reference loop when the container gets closed, but this can probably be addressed in a separate PR

mikeboers: I've been looking into overriding the `tp_traverse` and `tp_clear` so we can have cycles with `__del__` and still deal with it. Not sure if that is something we can actually do, but...

mikeboers: I think we could leave the `ContainerProxy` as a sort of `ContainerPointer` that only deals with the de-alloc, which would make the code a LOT simpler... if we removed `Stream.seek`.

(Of the top of my head at least).

jlaine:
> Yeah... they all do. That was my biggest holdup. But I figure if I flake8 them first, it won't be too bad.

Agreed. Let's get this PR merged next as some others will need refactoring afterwards.

mikeboers:
Just pushed the rebase. That was a LOT easier than I thought it would be. I'm now quite happy we pulled the trigger on that flake8 stuff.

PR #512, opened 2019-04-28
Make input/output buffer size configurable

mephi42:
This is useful when working with file-like objects that go to
Amazon S3, because it makes it possible to reduce the amount of
requests.

### DIFF
--- a/av/container/core.pyx
+++ b/av/container/core.pyx
@@ -32,7 +32,8 @@ cdef object _cinit_sentinel = object()
 
 cdef class Container(object):
 
-    def __cinit__(self, sentinel, file_, format_name, options, container_options, stream_options, metadata_encoding, metadata_errors):
+    def __cinit__(self, sentinel, file_, format_name, options, container_options, stream_options, metadata_encoding, metadata_errors,
+                  buffer_size):
 
         if sentinel is not _cinit_sentinel:
             raise RuntimeError('cannot construct base Container')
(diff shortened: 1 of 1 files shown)


av/container/core.pyx
```diff
@@ -201,6 +203,7 @@ def open(file, mode=None, format=None, options=None,
         reading on Python 2 (returning ``str`` instead of ``unicode``).
     :param str metadata_errors: Specifies how to handle encoding errors; behaves like
         ``str.encode`` parameter. Defaults to strict.
+    :param int bufsize: Decoder buffer size. Defaults to 32k.
```
jlaine: I think a better description is needed for the parameter, if it is used both for input and output streams I'm guessing "Decoder buffer size" isn't quite accurate? Also it would be nice to state what unit it is in.. bytes presumably?

mephi42: Thanks, I'll expand the description.

`buffer_size` sounds good, because it matches the name of the corresponding `avio_alloc_context` parameter.

mikeboers:
Looks good to me!

My regret is that the current setup with `av.open` constructing the containers is that you can't modify the containers before they start their work, and so we have to keep passing more and more things via `av.open`. But... that is another deal entirely.

If others are happy, I'd like to see this merged. @jlaine if you're already rebasing a pile of things for the Travis PR, and it isn't any added trouble, please feel free to grab this. Otherwise I'll wait for the dust to settle and do it.

mephi42:
Yes, it's only used when we pass a file-like object to `av.open()`. I should mention that in the description as well.

It looks as if `max_packet_size` was used to restrict the size of reads here https://github.com/FFmpeg/FFmpeg/blob/bf5d814f1082ea330ba78a591eb426c6c4263c0c/libavformat/aviobuf.c#L551, but this appears to be no longer the case.

I also noticed that `self.bufsize` is not necessary anymore, but kept it for compatibility.

jlaine:
Thank you @mephi42 !

PR #513, opened 2019-05-01
Make it possible to close input container

jlaine:
This is useful to explicitly close resources such as webcams, without having to wait for the GC to kick in.


mikeboers:
Looks great. Make it go.

PR #514, opened 2019-05-02
Stop using nose to run tests

jlaine:
Vanilla "unittest" is good enough since Python 2.7 / 3.2.


mikeboers:
I'm cool with this change. Hopefully Travis likes it more.

jlaine:
Thanks for the feedback @mikeboers !

I'll wait for CI to complete before merging then I'll rebase the appveyor / 3.7 PR.

An interesting point: there was some logging output I didn't previously see when using nose. It would seem recent ffmpeg versions have eliminated a number of warnings.

PR #519, opened 2019-05-06
Rework enums to be PyPy-compatible

jlaine:


jlaine:
OK, well this broke some things on Python 3.4 as dicts are not ordered.

jlaine:
PyPy 2.7 fully passes the test suite! PyPy 3.5 and 3.6 have two failures, because Travis's PyPy versions don't seem to work correctly with numpy, the multiarray module fails to import.

tests/test_enums.py
```diff
@@ -21,8 +21,6 @@ def test_basics(self):
 
         cls = self.define_foobar()
 
-        self.assertIsInstance(cls, EnumType)
```
mikeboers: Does this not still work?

mikeboers:
I pushed a couple small changes to your branch. Feel free to disregard them if you disagree.

I just thought it would be keen to maintain the normal `type.__new__` API.

Did you happen to do a version at some point where EnumType was still cdef? I feel like that would still work. Although, I could/should not care because that wouldn't affect the use of the enums in a significant way.

tests/test_enums.py
mikeboers: It does in my version. The commits pushed below have this. Sorry if this breaks CI. 😉

PR #528, opened 2019-05-20
Detect static library and pthread awareness in setup.py

stepheny:
It is great to stop trying to build against static ffmpeg library early. But only showing "unknown flags" -pthread is very confusing.
This patch trys to check `'library_dirs'` returned by `parse_cflags`.
If system ffmpeg was used, this should be empty. If a custom built ffmpeg was used, probably in /usr/local/lib, this would be filled by `pkg-config`. And then check if target `.so` file exists there. If it does not exist, the library is probably static. In case this test could go wrong, an env var was added to bypass the test.
I think this should work on linux. I've no idea how this would affect windows or macos users.
I've no idea whether the env var name and error message are proper.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,10 +63,14 @@ _cflag_parser.add_argument('-L', dest='library_dirs', action='append')
 _cflag_parser.add_argument('-l', dest='libraries', action='append')
 _cflag_parser.add_argument('-D', dest='define_macros', action='append')
 _cflag_parser.add_argument('-R', dest='runtime_library_dirs', action='append')
+_cflag_parser.add_argument('-pthread', dest='pthread', action='store_true')
 def parse_cflags(raw_cflags):
     raw_args = shlex.split(raw_cflags.strip())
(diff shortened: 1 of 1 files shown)


mikeboers:
This makes me nervous. I'd really like to have something that is aware of the static vs. dynamic build problem, but I'd rather the failure case be that it keeps compiling.

On the other hand, as long as there is a big message about why it is stopping the build, and we ask for feedback here when it make a false negative, I think that could be quite reasonable.

Here we're covering `.so` and `.dylib`. Neither @jlaine or I use Windows so can anyone out there confirm that this looks vaguely reasonable there?

jlaine:
I went with a much simpler approach in 4b56c73c33e84fd0a9105053e201c6913d74377b : if `-pthread` is found in the flags we display:

```
pkg-config returned flags we don't understand: -pthread -pthread -pthread
Building PyAV against static FFmpeg libraries is not supported.
``

PR #532, opened 2019-06-08
Fix a typo in README

gemfield:


mikeboers:
I wasn't keen on the comment, so I just committed something like this that is slightly more clarifying. Thanks!

PR #540, opened 2019-07-15
Update context.pyx

### LINKED ISSUE #340: av.bytesource.Bytesource object has no attribute 'size' Error while adding extradata manually.
HI. 

I'm trying to decode h.264 raw stream without PPS SPS. 

For encoding parameters, some extradata is being delivered from another session.

But while trying to add extradata bytes manually, the script returns the error below.

line 392, in initialize_ffmpeg_video
    av_codec_ctx.extradata = extradata
  File "av/codec/context.pyx", line 103, in av.codec.context.CodecContext.extradata.__set__
**AttributeError: 'av.bytesource.ByteSource' object has no attribute 'size'**

this code block in av/codec/context.pyx causes the issue.

image

I have changed the line that is allocating extradata_size to

    self.ptr.extradata_size = len(data)

then, it works.

in bytesource.pyx, there is no attribute 'size' for bytesource object. 

The original commit from issue #287 by @adavoudi includes the working code. 
(commit 51a7385710ae2973ed5daa3f877434226d696645 on 25 Feb)

       def __set__(self, data):
           buffer_size = len(data)
           cdef ByteSource source = bytesource(data)
           self.ptr.extradata =  <uint8_t *>malloc(buffer_size * sizeof(uint8_t))
           self.ptr.extradata_size = buffer_size
           memcpy(self.ptr.extradata, source.ptr, buffer_size)

but it has been changed in merged commit 4b01b63 on 23 Mar.

please check it out. 

Thanks.

ramoncaldeira:
Fixes #340

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -96,7 +96,7 @@ cdef class CodecContext(object):
         def __set__(self, data):
             self.extradata_source = bytesource(data)
             self.ptr.extradata = self.extradata_source.ptr
-            self.ptr.extradata_size = self.extradata_source.size
+            self.ptr.extradata_size = self.extradata_source.length
 
     property extradata_size:
         def __get__(self):


jlaine:
CI failures are not your fault, we need to fix them.

I'm starting to seriously hate Travis, the failure is way up at the top, it (silently) complains "brew is out of date", yet doesn't error immediately.

I had the same problem a few days ago, the "fix" is to force an update of Homebrew, which adds several minutes to the build time and nullifies the benefit of the brew "add-on":

https://github.com/aiortc/aiortc/commit/a15a2c98c1cc223ab467314fc92b5cc4b1ff6149

mikeboers:
Cherry-picked.

Thanks for this and your patience.

Issue #545, opened 2019-08-26
Support sub-interpreters


mikeboers:
From Gitter:

> Newby question... is PyAV compatible with python sub-interpreters? According to the python docs https://docs.python.org/3/c-api/init.html using `PyGILState_Ensure()`, for example, is not compatible and I see this in the generated `.c` files inside `#ifdef WITH_THREAD`. I am trying to use PyAV inside a Django wsgi app and running into intermittent hang problems. Thanks in advance for your help.

I see that `PyGILState_Ensure` is called by `__Pyx_RefNannySetupContext` in every file. There are only 4 places that is called with acquire_gil being non-zero:

```
src/av/container/pyio.c: __Pyx_RefNannySetupContext("pyio_read", 1);
src/av/container/pyio.c: __Pyx_RefNannySetupContext("pyio_write", 1);
src/av/container/pyio.c: __Pyx_RefNannySetupContext("pyio_seek", 1);
src/av/logging.c: __Pyx_RefNannySetupContext("log_callback", 1);
```

So... if you don't use Python based IO, and disable the logging system, that should be clear.

[This triggers it](https://github.com/mikeboers/PyAV/blob/develop/av/video/frame.pyx#L103-L104) in each of the input/output containers, but I don't think control reaches it. Same with utils. [This is another `with gil` in logging](https://github.com/mikeboers/PyAV/blob/develop/av/logging.pyx#L229).

Okay... so there isn't a user-exposed way to disable the logging system adequately enough. We would have to restructure the `Frame._init` method just a bit to avoid the `with gil`, and add another method to logging to disable it entirely (or maybe check an envvar or something).

At that point, I don't think it will get called (if you avoid Python IO)

mikeboers:
Continued in #546.

PR #546, opened 2019-08-27
Subinterpreters

mikeboers:
Users can disable the logging system, which is the harder one the avoid.


mikeboers:
Continuing here instead of in #545.

It seems that Cython 0.29 explicitly checks for and bails in sub-interpreters. The latest PyPI distribution was built with Cython 0.29. Given that, I don't know how much point there is in this.

PR #549, opened 2019-09-01
One line change to make CodecContext.parse use bytes instead of str

philkr:
Avoids encoding and decoding the raw data.

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -167,7 +167,7 @@ cdef class CodecContext(object):
             id(self),
         )
 
-    def parse(self, str input_, allow_stream=False):
+    def parse(self, bytes input_, allow_stream=False):
 
         if not self.parser:
             self.parser = lib.av_parser_init(self.codec.ptr.id)


jlaine:
Unless I'm mistaken this "one-line-change" is actually a breaking API change?

philkr:
thank your for the quick reply. I just briefly checked the library and didn't see it used anywhere, except for in the scratchpad [cctx_decode.py](https://github.com/mikeboers/PyAV/blob/develop/scratchpad/cctx_decode.py).

Do you think there is a better way for fixing this? Maybe accept either str or bytes?

jlaine:
> thank your for the quick reply. I just briefly checked the library and didn't see it used anywhere, except for in the scratchpad [cctx_decode.py](https://github.com/mikeboers/PyAV/blob/develop/scratchpad/cctx_decode.py).

I'll let @mikeboers weigh in on this as I honestly don't know how it's used.

> Do you think there is a better way for fixing this? Maybe accept either str or bytes?

I don't think it should be one of those "ambiguous" APIs, there is most likely one "right" type. What are we parsing here?

mikeboers:
Bytes is almost certainly the way to go.

I doubt anyone is really using this, and I'm okay with this being a breaking change because it is so wrong.

jlaine:
@philkr could you update your PR please to fix the scratchpad entry you spotted so that it opens the file in binary mode?

I'll try to find some time to get CI back on its feet.

mikeboers:
I'll make the update. I'm merging it now.

Re Travis, develop will now be updating homebrew, which is dumb, but required AFAICT.

mikeboers:
Merged via cherry-pick onto new develop.

jlaine:
> Re Travis, develop will now be updating homebrew, which is dumb, but required AFAICT.

I'm falling out of love with Travis, fast :(

The homebrew integration is abysmal and shows no signs of improving.

PR #550, opened 2019-09-02
Don't require macOS with FFmpeg < 4.0 to pass.

mikeboers:
We don't really care enough to figure this out.


mikeboers:
Cherry-picked.

PR #557, opened 2019-10-07
Add support for extracting motion vectors

RouquinBlanc:
There's a lot of fun to be made using motion vectors, even more from python side!

Please let me know if the approach is correct, and what would be missing in order to integrate this feature.


mikeboers:
Howdy!

I wasn't too keen on how this felt like it was specific for you needs, so I took it and built a more general side data feature over in #558.

Thanks for putting this together. It really helped me understand this and I think we will have a more flexible and general system now.

PR #558, opened 2019-10-16
Expose Frame Sidedata

mikeboers:
I wasn't too keen on #557 (because it appeared to be tailored to a specific use-case), but it did show me how to work with side data, so I whipped this together.

Thoughts, anyone?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,6 @@ deploy-docs: docs
 
 
 
-clean: clean-build
 
 clean-build:
 	- rm -rf build
@@ -87,4 +86,5 @@ clean-docs:
 	- rm tmp/tagfile.xml
 	- make -C docs clean
 
+clean: clean-build clean-sandbox clean-src
 clean-all: clean-build clean-sandbox clean-src clean-docs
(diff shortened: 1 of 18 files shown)


RouquinBlanc:
Hi Mike!

Indeed, this looks much more generic than my crafty attempt at Cython! But you got my point and I'm very happy to see that integrated, thanks a lot! I will give it a try ASAP

RouquinBlanc:
I can confirm getting similar results with the C equivalent. And also still being able to get the whole list as an ndarray is much appreciated! 👍

mikeboers:
The way I padded out the array is kinda weird. Why does `sizeof(AVMotionVector)` not equal what it looks like it should? I feel like that needs to be figured out, or generalized in some way.

mikeboers:
[`numpy.dtype` has an `align` parameter](https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.html#numpy.dtype) which is what I need. Hooray!

rohit-gupta:
Being able to extract motion vectors without writing C code would be a great feature !

PR #564, opened 2019-10-26
Expose AVStream.r_frame_rate and av_guess_frame_rate()

mikeboers:
This is continuing from #482.


mikeboers:
Does `base_frame_rate` and `guessed_frame_rate` sound good for these two?

KengoSawa2:
I think guessed_frame_rate is good:)
it's less confusing than base_frame_rate

PR #566, opened 2019-10-26
Add VideoFrame.save(path, ...) method.

mikeboers:


WyattBlue:
What's the use case here?

Issue #577, opened 2019-12-09
Allow different interpolation to VideoFrame.reformat


legraphista:
## Overview

There is no way to scale a video frame with other interpolation methods.
`SWS_BILINEAR` is the best fit as a default, but for speed you would want `SWS_FAST_BILINEAR`, and `SWS_POINT`(neighbor) for up-scaling small frames (64x64) to better visualize them and keeping the information clean.

Using OpenCV would be an option but there are too many hoops to jump through. It's way faster to do in the native libav API.

## Desired Behavior

Ability to use different interpolation for resizing a video frame. Right now it's hardcoded to use `SWS_BILINEAR` 

## Example API

```python
frame.reformat(width=width,
               height=height,
               interpolation='neighbor')
# or maybe even better
frame.reformat(width=width,
               height=height,
               interpolation=av.video.interpolation.NEIGHBOR)
```

## Additional context

Location of hardcoded flags:
https://github.com/mikeboers/PyAV/blob/ae64e1ba52bea3cb72cfec473233f5137d0c1bda/av/video/frame.pyx#L200-L212

mikeboers:
That's a keen idea, and a small thing to add.

PR #579, opened 2019-12-10
Expose swscale flags

stepheny:
Early implement for #577 
I've been using this indoor for months. My main hesitation to upstream this was its change to API.
A new argument with default value was added to `VideoFrame._reformat`. It seems that argument with default value was never use in `cdef` in this project, but this is probably the only way not to break existing code.
Please consider this as a proposal. I would like to update this PR to meet up with an updated API design.


mikeboers:
Thanks for this. I did a bit of work to style it the same as the rest of our constants, and then refactored the reformatting into the actual `VideoReformatter` class.

It should now be in develop.

PR #581, opened 2019-12-12
updated examples, added container.close()

mastak:
I took code from example as a base and got a memory leak. Reason: there is no `close` method for the container.

I verified on small video: 2sec, 360:

without close:
without_close

with close:
with_close

Simplified code:
```python
import av
import av.datasets
import psutil

current_process = psutil.Process()

get_mem_uss = lambda: current_process.memory_full_info().uss

PATH = '...'


def read_frames(path):
    container = av.open(path)
    return [frame.to_ndarray(format='yuv420p') for frame in container.decode(video=0)]


started_with = get_mem_uss()
for i in range(15000):
    read_frames(PATH)
    if i % 50:
        print(f'Used memory {(get_mem_uss() - started_with) / 1024 / 1024}')

```

### DIFF
--- a/examples/basics/remux.py
+++ b/examples/basics/remux.py
@@ -1,4 +1,3 @@
-import av
 import av.datasets
 
 
@@ -23,4 +22,5 @@ for packet in input_.demux(in_stream):
 
     output.mux(packet)
 
+input_.close()
 output.close()
--- a/examples/basics/save_keyframes.py
+++ b/examples/basics/save_keyframes.py
@@ -1,20 +1,18 @@
-import av
 import av.datasets
 
 
-container = av.open(av.datasets.curated('pexels/time-lapse-video-of-night-sky-857195.mp4'))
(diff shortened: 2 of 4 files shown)


mikeboers:
Can you confirm for me that I'm reading the first one right? It looks like it builds until shortly after 4.5k, then takes 0.7s and cleans most of the memory up.

I just want to fix that behaviour first. But instead, could update the docs to talk about the current behaviour, and then I can remove it later if/when it is dealt with. 😛

mikeboers:
I cherry picked and merged.

mastak:
> Can you confirm for me that I'm reading the first one right? It looks like it builds until shortly after 4.5k, then takes 0.7s and cleans most of the memory up.

Yes. Firstly it looks like memory leak but after ~5к iterations it clean memory with a spike at CPU.

mikeboers:
Cool. It is overwhelmingly likely the Container to Stream cycle. I need to clean that up at some point.

PR #583, opened 2019-12-19
added yuvj420p format for to_ndarray

mastak:
`yuvj420p` has the same shape as `yuv420p`. The difference in color ranges: 

- `yuv420p` according to rec 709 (https://en.wikipedia.org/wiki/Rec._709#Digital_representation) has a nominal range of [16..235] for Y channel and CB and CR channels have a nominal range of [16..240]. rec 601 has limited ranges also. It's main recommendations for `yuv420p` (https://en.wikipedia.org/wiki/YUV) 

- `yuvj420p` like I420, but with a full range ("digital", 0-255) (https://wiki.videolan.org/YUV#J420)

### DIFF
--- a/av/video/frame.pyx
+++ b/av/video/frame.pyx
@@ -236,7 +236,7 @@ cdef class VideoFrame(Frame):
 
         import numpy as np
 
-        if frame.format.name == 'yuv420p':
+        if frame.format.name in ('yuv420p', 'yuvj420p'):
             assert frame.width % 2 == 0
             assert frame.height % 2 == 0
             return np.hstack((
@@ -277,7 +277,7 @@ cdef class VideoFrame(Frame):
         """
         Construct a frame from a numpy array.
         """
-        if format == 'yuv420p':
+        if format in ('yuv420p', 'yuvj420p'):
             assert array.dtype == 'uint8'
             assert array.ndim == 2
             assert array.shape[0] % 3 == 0
--- a/tests/test_videoframe.py
+++ b/tests/test_videoframe.py
@@ -236,6 +236,14 @@ class TestVideoFrameNdarray(TestCase):
         self.assertEqual(frame.format.name, 'yuv420p')
         self.assertTrue((frame.to_ndarray() == array).all())
 
+    def test_ndarray_yuvj420p(self):
+        array = numpy.random.randint(0, 256, size=(720, 640), dtype=numpy.uint8)
+        frame = VideoFrame.from_ndarray(array, format='yuvj420p')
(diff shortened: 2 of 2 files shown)


av/video/frame.pyx
```diff
@@ -236,7 +236,7 @@ cdef class VideoFrame(Frame):
 
         import numpy as np
 
-        if frame.format.name == 'yuv420p':
+        if frame.format.name in ('yuv420p', 'yuvj420p'):
```
mikeboers: What does the `j` mean that doesn't affect the shape?

mastak: `yuvj420p` has the same shape as `yuv420p`. The difference in color ranges: 

- `yuv420p` according to rec 709 (https://en.wikipedia.org/wiki/Rec._709#Digital_representation) has a nominal range of [16..235] for Y channel and CB and CR channels have a nominal range of [16..240]. rec 601 has limited ranges also. It's main recommendations for `yuv420p` (https://en.wikipedia.org/wiki/YUV) 


- `yuvj420p` like I420, but with a full range ("digital", 0-255) (https://wiki.videolan.org/YUV#J420)

mastak:
Script with comparing `pyav` vs native ffmpeg run as subprocess:

requirements:

Install imageio-ffmpeg, for run `ffmpeg` as a subprocess and read raw data (https://github.com/imageio/imageio-ffmpeg/blob/master/imageio_ffmpeg/_io.py#L63)
```bash
pip install git+https://github.com/imageio/imageio-ffmpeg.git@53b1533382ec6502e2e02a3640898603a03fc00f
```

```python
import av
import numpy as np
from imageio_ffmpeg import read_frames
from av.datasets import fate


def av_arrays(path_):
    return [frame.to_ndarray(format='yuvj420p') for frame in av.open(path_).decode(video=0)]


def ffmpeg_arrays(path_):
    gen = read_frames(path=path_, pix_fmt='yuvj420p', bits_per_pixel=12)
    meta = next(gen)
    data = list(gen)
    yuv = np.frombuffer(b''.join(data), dtype='uint8')
    w, h = meta['size']
    return yuv.reshape(-1, int(h * 1.5), w)

path = fate('h264/interlaced_crop.mp4')
av_frames = av_arrays(path)
ffmpeg_frames = ffmpeg_arrays(path)

for av_frame, ffmpeg_frame in zip(av_frames, ffmpeg_frames):
    assert np.array_equal(av_frame, ffmpeg_frame)

```

jlaine:
This looks good to, let's go!

PR #601, opened 2020-01-23
Support for more pixel formats

mildsunrise:
- Fixes plane count calculation at `VideoFrame.planes` (in cases when many components are stored in the same plane).
- Corrects the `VideoPlane` constructor so that the palette plane in `pal8` is accessible.
- Implements ndarray conversion with `rgb8` / `bgr8`.
- Implements ndarray conversion with `pal8` (by accepting a `palette` argument / returning a tuple).

### DIFF
--- a/av/video/frame.pyx
+++ b/av/video/frame.pyx
@@ -132,7 +132,11 @@ cdef class VideoFrame(Frame):
         # ourselves to the maximum plane count (as determined only by VideoFrames
         # so far), in case the library implementation does not set the last
         # plane to NULL.
-        cdef int max_plane_count = self.format.ptr.nb_components
+        cdef int max_plane_count = 0
+        for i in range(self.format.ptr.nb_components):
+            count = self.format.ptr.comp[i].plane + 1
+            if max_plane_count < count:
+                max_plane_count = count
         cdef int plane_count = 0
         while plane_count < max_plane_count and self.ptr.extended_data[plane_count]:
             plane_count += 1
@@ -257,7 +261,7 @@ cdef class VideoFrame(Frame):
             return useful_array(frame.planes[0], 3).reshape(frame.height, frame.width, -1)
         elif frame.format.name in ('argb', 'rgba', 'abgr', 'bgra'):
             return useful_array(frame.planes[0], 4).reshape(frame.height, frame.width, -1)
-        elif frame.format.name in ('gray', 'gray8'):
+        elif frame.format.name in ('gray', 'gray8', 'rgb8', 'bgr8'):
             return useful_array(frame.planes[0]).reshape(frame.height, frame.width)
         else:
             raise ValueError('Conversion to numpy array with format `%s` is not yet supported' % frame.format.name)
@@ -309,7 +313,7 @@ cdef class VideoFrame(Frame):
             assert array.dtype == 'uint8'
             assert array.ndim == 3
             assert array.shape[2] == 4
-        elif format in ('gray', 'gray8'):
+        elif format in ('gray', 'gray8', 'rgb8', 'bgr8'):
             assert array.dtype == 'uint8'
             assert array.ndim == 2
         else:
--- a/av/video/frame.pyx
+++ b/av/video/frame.pyx
@@ -137,6 +137,9 @@ cdef class VideoFrame(Frame):
             count = self.format.ptr.comp[i].plane + 1
             if max_plane_count < count:
                 max_plane_count = count
+        if self.format.name == 'pal8':
+            max_plane_count = 2
+
         cdef int plane_count = 0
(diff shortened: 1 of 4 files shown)


jlaine:
Could you please rebase this on top of master and provide some unit tests for the new formats as is done with the existing ones?

I see there is a breaking API change, could you please explain why?

mildsunrise:
The only breaking change is that a `VideoFrame` no longer has a `component` property (because it's an incorrect concept; a video frame may have more than one component, or may not be associated with a component at all).

Does this seem okay? If it doesn't, I can remove the breaking change, but I strongly think it's incorrect for this property to be there.

jlaine:
I'll let @mikeboers weigh in on the breaking change but I understand the reasoning.

In the meantime can you please add the tests I asked for? Format handling is a tricky business and I want to make sure the behaviour is locked down by tests.

A recent commit for pixel formats which included a test : 028b74225ea3b1c20695a7ceeb675c9e0ad586c1

mildsunrise:
Sure, I'll rebase and add a test for `rgb8` and another one for `pal8` when I have some time.

Also, I think it would be more consistent for `from_ndarray` to accept an `(image, palette)` tuple for pal8, rather than a `palette` keyword argument...

av/video/frame.pyx
```diff
@@ -282,10 +292,24 @@ cdef class VideoFrame(Frame):
...
+        if format == 'pal8':
+            assert array.dtype == 'uint8'
+            assert array.ndim == 2
+            assert palette.dtype == 'uint8'
+            assert palette.shape == (256, 4)
+            frame = VideoFrame(array.shape[1], array.shape[0], format)
+            copy_array_to_plane(array, frame.planes[0], 1)
+            frame.planes[1].update(palette.view('>i4').astype('i4').tobytes())
+            return frame
+        assert palette is None
```
mikeboers: Can this be a `ValueError` with a message please?

av/video/plane.pxd
```diff
@@ -7,3 +7,4 @@ cdef class VideoPlane(Plane):
     cdef VideoFormatComponent component
```
mikeboers: This can be removed.

mikeboers:
I whole-heartily approve of a breaking change re. fixing the erroneous notion that a `VideoPlane` has a single component.

I don't think I've personally ever used the components on planes, so I don't know how much of an authority I can be on what is the best of:

- no component(s) on VideoPlane;
- `primary_component` for the first one;
- `components` as a tuple of them all.

(...although my gut says the last one TBH.)

I'm leaving a few specific notes in places.

Finally... thanks for this! Sorry I'm a bit of an absentee maintainer lately.

av/video/frame.pyx
```diff
@@ -137,6 +137,9 @@ cdef class VideoFrame(Frame):
             count = self.format.ptr.comp[i].plane + 1
             if max_plane_count < count:
                 max_plane_count = count
+        if self.format.name == 'pal8':
+            max_plane_count = 2
+
```
mikeboers: So... does `pal8` actually have 2 planes, or are we (ab)using `VideoPlane` as a way to expose the palette?

mildsunrise: `pal8` does have two planes, the second one is for storing the palette

av/video/plane.pxd
mildsunrise: Done!

av/video/frame.pyx
mildsunrise: I've applied the change mentioned in https://github.com/mikeboers/PyAV/pull/601#issuecomment-589947756, which removes the need for an extra parameter and makes the API more consistent.

av/video/frame.pyx
mikeboers: And the code above it won't result in `max_plane_count == 2`? That seems weird. Okay.

mikeboers:
Whenever y'all are happy, I'm happy. @jlaine feel free to take whatever action if you want.

av/video/frame.pyx
mildsunrise: It won't because the palette has no associated component, it's just metadata.

mildsunrise:
I've rebased and added tests for `rgb8`, `bgr8` and `pal8`. As for this:

> `components` as a tuple of them all.

This looks better, but there's still planes (like the palette one) that are not referenced by any component, because they're metadata. I think it's better to not complicate the exposed API, but I'm happy either way.

jlaine:
This all looks very neat and tidy to me. Not sure why I don't have the "merge" button, maybe yet another rebase neede?

mildsunrise:
strange, there are no conflicts :thinking: I have rebased anyway, let me know if it helps

mikeboers:
Shall we roll with it as is? Don't need to hold up because Mike wants to over-engineer everything...

PR #611, opened 2020-03-14
[build] remove PYAV_LIBRARY_NAME

jlaine:
The library name is always ffmpeg nowadays.

### DIFF
--- a/scripts/build-deps
+++ b/scripts/build-deps
@@ -30,7 +30,7 @@ cd "$PYAV_LIBRARY_ROOT"
 
 # Download and expand the source.
 if [[ ! -d $PYAV_LIBRARY ]]; then
-    url="https://$PYAV_LIBRARY_NAME.org/releases/$PYAV_LIBRARY.tar.gz"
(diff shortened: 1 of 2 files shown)


scripts/build-deps
```diff
@@ -40,18 +40,16 @@ fi
...
-    --disable-optimizations \
     --disable-mmx \
+    --disable-optimizations \
+    --disable-static \
+    --disable-stripping \
     --enable-debug=3 \
     --enable-gpl \
     --enable-libx264 \
+    --enable-shared \
     $CONFFLAGS \
```
mikeboers: Does $CONFFLAGS do anything? If not, either remove it or call it `PYAV_LIBRARY_CONFIG` or something. (Better to just remove since this is for testing.)

mikeboers: That came off rather commanding. Sorry. Was just writing notes, not trying to issue orders.

jlaine: No worries, the first comment was "I love it" ;)

PR #620, opened 2020-04-13
Improve pixel format handling

mildsunrise:
### Allow user to set preferred format

Expose [`get_format`](https://libav.org/documentation/doxygen/master/structAVCodecContext.html#a0d8f46461754e8abea0847dcbc41b956) to allow user to choose decoded pixel format, by adding a `VideoCodecContext.preferred_format` property. If set and supported by the decoder, that format will be used; otherwise fall back to current behaviour.

We could have exposed the callback directly for better control, but since there's `codec.video_formats` and PyAV tries to hide these complexities, I thought it was better this way. WDYT?

### Handle format overrides, not set

The current code only updates `VideoCodecContext.format` and `VideoCodecContext.pix_fmt` when the user modifies them. However, according to the documentation, decoders may override those values too (at open, when decoding frames, etc.). The code now checks for `pix_fmt` / `width` / `height` changes at every access.

Also, when the pixel format is not set, `pix_fmt` now returns None instead of raising.

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -157,6 +157,7 @@ cdef class CodecContext(object):
     cdef _init(self, lib.AVCodecContext *ptr, const lib.AVCodec *codec):
 
         self.ptr = ptr
+        self.ptr.opaque = <void*>self
         if self.ptr.codec and codec and self.ptr.codec != codec:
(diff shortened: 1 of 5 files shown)


jlaine:
This has remained a draft for 2 years, closing. Feel free to resubmit a fresh PR.

PR #631, opened 2020-04-22
build source package and binary wheels from CI

jlaine:


jlaine:
@mikeboers we got our first batch of Linux wheels :)

https://github.com/mikeboers/PyAV/actions/runs/85144566

image

jlaine:
.. and now Mac too!

mikeboers:
That is a lot of wheels.

Issue #600, opened 2020-01-22
outcontainer unicodeencodeerror


woshimanong1990:
**IMPORTANT:** Be sure to replace all template sections {{ like this }} or your issue may be discarded.

## Overview

the out file path contain chinese character, it report  unicodeencodeerror

I have fix the bug. But I don't know how to sumbit a PR

`
 
        diff --git a/av/container/output.pyx b/av/container/output.pyx
        index 8781958..daefd1c 100644
        --- a/av/container/output.pyx
        +++ b/av/container/output.pyx
        @@ -1,5 +1,6 @@
        from fractions import Fraction
        import logging
        +import sys

        from av.codec.codec cimport Codec
        from av.container.streams cimport StreamContainer
        @@ -11,6 +12,14 @@ from av.utils cimport dict_to_avdict

        from av.dictionary import Dictionary

        +try:
        +    from os import fsencode
        +except ImportError:
        +    _fsencoding = sys.getfilesystemencoding()
        +
        +    def fsencode(s):
        +        return s.encode(_fsencoding)
        +

        log = logging.getLogger(__name__)

        @@ -162,7 +171,9 @@ cdef class OutputContainer(Container):
                    stream._finalize_for_output()

                # Open the output file, if needed.
        -        cdef char *name = "" if self.file is not None else self.name
        +        file_name = "" if self.file is not None else self.name
        +        cdef bytes name_obj = fsencode(file_name) if isinstance(file_name, unicode) else file_name
        +        cdef char *name = name_obj
                if self.ptr.pb == NULL and not self.ptr.oformat.flags & lib.AVFMT_NOFILE:
                    err_check(lib.avio_open(&self.ptr.pb, name, lib.AVIO_FLAG_WRITE))

`

mikeboers:
Thanks! I'll make this change at some point. =]

necan:
seems not fixed yet

jlaine:
Hi @necan could you give me an example of a filename which triggers this bug please so that we can capture it in a unit test?

Now that we only support Python 3.5+ we can unconditionally use `os.fsencode()`

necan:
@jlaine it appears in Windows with Simplified Chinese Locale as it uses GB18030 as the default character encoding of cmd or Powershelll.

jlaine:
Could you give PR #643 a try and let me know if it fixes the issue for you?

necan:
@jlaine build error

```
    Running setup.py install for av ... error
[...]

PR #643, opened 2020-04-29
Use os.fsencode for both input and output file names (fixes: #600)

jlaine:
It's safe to call os.fsencode() regardless of whether the filename was
passed as `bytes` or `str`. Remove shim for `os.fsencode()`, it has been
available since Python 3.2.


jlaine:
@mikeboers does this look good to you?

mikeboers:
Looks good to me!

PR #645, opened 2020-04-29
Remove some references to Python 2

jlaine:
Also simplify the `dict_to_avdict` signature as we:

- always pass `clear=True`
- always explicitly pass the `encoding` and `errors` arguments

### DIFF
--- a/av/error.pyx
+++ b/av/error.pyx
@@ -41,12 +41,11 @@ cpdef tag_to_code(bytes tag):
     """
     if len(tag) != 4:
         raise ValueError("Error tags are 4 bytes.")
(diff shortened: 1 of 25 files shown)


av/error.pyx
```diff
@@ -41,12 +41,11 @@ cpdef tag_to_code(bytes tag):
     """
     if len(tag) != 4:
         raise ValueError("Error tags are 4 bytes.")
-    atag = bytearray(tag)  # Python 2. Ugh.
```
mikeboers, line 44: Ahaha.

mikeboers:
I love it.

PR #646, opened 2020-04-30
Remove support for Python 2 buffer protocol.

jlaine:


mikeboers:
Been long enough that those have been deprecated? 😛

jlaine:
Yeah maybe I should not have done this in the same PR but there are two distinct changes here:

- remove all the (unreachable) code for Python 2 buffer protocol

- remove the methods which have been deprecated since PyAV 0.5.3 (the one right before 6.0.0)

mikeboers:
There was a reason for the `_Buffer`, but I don't recall what it was. I guess it is no longer needed (or never was).

NBD about it being in the same PR. I'm not very clean myself.

jlaine:
Alrighty, here goes then!

Issue #640, opened 2020-04-28
Double freeing CodecContext.extradata?


mikeboers:
See [this comment](https://gitter.im/mikeboers/PyAV?at=5ea1dba61e3d5e20633cd8a5)

jlaine:
Yeah the memory ownerships looks very dodgy here. A `memcpy` seems like a reasonable idea.

haohuanw:
https://github.com/PyAV-Org/PyAV/pull/658#issue-412365032
pr for fixing this issue; I have applied this patch and tested with my offline use case and it works well.

jlaine:
To the best of my knowledge this was fixed in #756

PR #658, opened 2020-05-02
fix double free on extradata source in codec context

haohuanw:
https://github.com/PyAV-Org/PyAV/issues/640#issue-608274450

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -231,8 +231,9 @@ cdef class CodecContext(object):
 
         def __set__(self, data):
             self.extradata_source = bytesource(data)
-            self.ptr.extradata = self.extradata_source.ptr
+            self.ptr.extradata = <uint8_t *>malloc(self.extradata_source.length * sizeof(uint8_t))
             self.ptr.extradata_size = self.extradata_source.length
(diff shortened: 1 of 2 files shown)


haohuanw:
@jlaine would you be able to take a quick look on this PR?

jlaine:
This looks like you're touching on a legitimate point, however doesn't the fix introduce a memory leak? When is this malloc'd memory freed?

jlaine:
I had a look at the FFmpeg docs and it looks a little more involved than this:

- encoding: Set/allocated/freed by libavcodec.
- decoding: Set/allocated/freed by user.

https://ffmpeg.org/doxygen/4.0/structAVCodecContext.html#abe964316aaaa61967b012efdcced79c4

I'm not convinced we should allow setting `extradata` at all for encoders. For decoders, however it looks like we do own the memory and need to both allocate and free it. When we do perform allocation, I believe we need to add `AV_INPUT_BUFFER_PADDING_SIZE` to the extradata size.

jlaine:
@haohuanw I provided an alternate approach in #756 could you let me know if it solves your problem?

jlaine:
Fixed in #756

PR #668, opened 2020-05-23
Metadata UnicodeDecodeError Fix

indyquiet:
Some movies fail to decode just because of bad metadatas.

``Traceback (most recent call last):``
``File "av\utils.pyx", line 14, in av.utils._decode``
``UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte``

**Examples :**
Ignoring UnicodeDecodeError bytes from : b'AVI-Mux GUI 1.17.8, Aug 30 2008  12:36:58\xff'
-> 'AVI-Mux GUI 1.17.8, Aug 30 2008  12:36:58'

**Proposed solution :**
Merely ignore bad bytes by overriding the errors flag (from 'strict' to 'ignore')
also print the original metadata as a warning

### DIFF
--- a/av/utils.pyx
+++ b/av/utils.pyx
@@ -11,7 +11,12 @@ cimport libav as lib
 # ====================
 
 cdef _decode(char *s, encoding, errors):
-    return (<bytes>s).decode(encoding, errors)
+    try:
+      decoded = (<bytes>s).decode(encoding, errors)
+    except UnicodeDecodeError:
+      print("Ignoring UnicodeDecodeError bytes from : %s" % (str(<bytes>s)))
+      decoded = (<bytes>s).decode(encoding, errors='ignore')
+    return decoded
 
 cdef bytes _encode(s, encoding, errors):
     return s.encode(encoding, errors)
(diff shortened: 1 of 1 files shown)


indyquiet:
**List of all the original tags in these videos**
`b'M'`
`b'title'`
`b'AVI-Mux GUI 1.17.8, Aug 30 2008  12:36:58\xff'`
-> Except UnicodeDecodeError
`b'encoder'`
`b''`
`b'JUNK'`

jlaine:
This isn't a fix, it's a hack. If you want to ignore errors, why not pass `metdata_errors="ignore"` to `av.open`

indyquiet:
You don't seem to be getting the point : a simple warning is better than a bold crash. Whenever unsure, just copy ffmpeg's behaviour.

jlaine:
Does `metadata_errors="ignore"` not solve your issue? So far failed to make a convincing case for the change.

mikeboers:
I'm gonna agree with @jlaine here quite thoroughly. In general, I believe it is better to error as hard and fast as possible so issues don't propagate down the line.

If we could be convinced that `metadata_errors="ignore"` is not suitable, then I could see some sort of conditional, non-default, override that allows this sort of behaviour. But the bar to convince me is going to be very high.

PR #672, opened 2020-05-26
[discussion] retrieve rtsp ntp time

ramoncaldeira:
Hi! 

I have a use case where I would benefit from retrieving the camera's NTP time from a RTSP container. FFMPEG contains structures capable of parsing this information, but unfortunately they are internal and are not available in the public API.

I made this small patch to solve my problem. It is not generic enough, as it hardcodes ntp time into demux return. I also had to include `vendor/ffmpeg-4.2` in include_dirs on setup.py.

My intention with this PR is to share the solution with you and hopely find out a way to change the code so that it can be incorporated into the project.

### DIFF
--- a/av/container/input.pyx
+++ b/av/container/input.pyx
@@ -1,4 +1,4 @@
-from libc.stdint cimport int64_t
+from libc.stdint cimport int64_t, uint32_t, int32_t
 from libc.stdlib cimport malloc, free
 
 from av.container.streams cimport StreamContainer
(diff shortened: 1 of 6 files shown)


ramoncaldeira:
Run failed because of `fatal error: libavformat/rtsp.h: No such file or directory`.
This was expected, due to the private structures.

I tested the code myself with this snippet and it worked:
```import av

input_ = av.open('RTSP_URL')

for (ntp, packet) in input_.demux(video=0):
    print('ntp', ntp)
    print('packet', packet)

**Comment by florisdesmedt** [comment]
I would like to use this code, since I also need the timestamp from the rtcp packets. I'm however not that experienced with installing pyav using compilation, and using the `make` command I run into the issue mentioned above `fatal error: libavformat/rtsp.h: No such file or directory`. And as I would expect, after the failed compilation I can not import av.

How to I get this code running? Do I have to alter the Makefile to ensure the missing file is found? Or do I have to install PyAv using a package manager (pip, conda) and it will be replaced using the `make`?

Thanks in advance

**Comment by HoBeom** [comment]
I used your branch to download the video stream with RTSP. The returned ntp value is as follows, how should I interpret it?
```
...
1627412725.1005974
1627412724.9732378
1627412725.1492114
...
```

ramoncaldeira:
@HoBeom It's time in seconds since epoch. You can use this converter: https://www.epochconverter.com/.

samyhocine:
A very useful commit, unfortunately it fails completing the checking part :/

jlaine:
I'm sorry but I don't see this PR going anywhere, it completely changes the API.

jimibit:
> Hi!
> 
> I have a use case where I would benefit from retrieving the camera's NTP time from a RTSP container. FFMPEG contains structures capable of parsing this information, but unfortunately they are internal and are not available in the public API.
> 
> I made this small patch to solve my problem. It is not generic enough, as it hardcodes ntp time into demux return. I also had to include `vendor/ffmpeg-4.2` in include_dirs on setup.py.
> 
> My intention with this PR is to share the solution with you and hopely find out a way to change the code so that it can be incorporated into the project.

Great !! 
Very very useful.
Does this patch work in Windows ? Is it up-to-date ?

Jim

PR #693, opened 2020-07-27
Fix for #689: raise an exception when there is no decoder available

brglng:
Fix crash.


jlaine:
Could you please add a test capturing this situation?

jlaine:
This is done with a RuntimeError in e9d87b01ee412ca50180416c5cdf3fb996964a4c

PR #694, opened 2020-07-28
Fix for files whose nb_streams changes during demuxing #692

brglng:
Fix for files whose `nb_streams` changes during demuxing/decoding.

### DIFF
--- a/av/container/input.pyx
+++ b/av/container/input.pyx
@@ -111,6 +111,7 @@ cdef class InputContainer(Container):
         # (and others).
         id(kwargs)
 
+
         streams = self.streams.get(*args, **kwargs)
 
         cdef bint *include_stream = <bint*>malloc(self.ptr.nb_streams * sizeof(bint))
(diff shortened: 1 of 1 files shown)


av/container/input.pyx
```diff
@@ -111,6 +111,7 @@ cdef class InputContainer(Container):
         # (and others).
         id(kwargs)
 
+
```
jlaine, line 114: Please remove this unrelated change

av/container/input.pyx
```diff
@@ -143,6 +136,19 @@ cdef class InputContainer(Container):
                 except EOFError:
                     break
 
+                if self.ptr.nb_streams != len(self.streams):
+                    self.streams = StreamContainer()
+                    for i in range(self.ptr.nb_streams):
+                        self.streams.add_stream(wrap_stream(self, self.ptr.streams[i]))
+
+                for i in range(self.ptr.nb_streams):
+                    include_stream[i] = False
```
jlaine, line 145: How do we know `include_stream` is the correct size, it only seems to be allocated once?

jlaine:
I'm not comfortable with merging this change without any kind of unit tests. Can you please provide tests which increase / reduce the number of streams to make sure we exercise all possible conditions?

av/container/input.pyx
brglng: I should re-allocate `include_stream`. Thanks.

brglng:
> I'm not comfortable with merging this change without any kind of unit tests. Can you please provide tests which increase / reduce the number of streams to make sure we exercise all possible conditions?

OK, I will add a unit test. Could you please tell me where I can upload the input file which is used to reproduce this issue?

PR #712, opened 2020-10-14
Cython3

KOLANICH:
Implemented support of Cython 3 and also replaced absolute imports with relative ones to make the package a bit more portable.

### DIFF
--- a/av/audio/format.pyx
+++ b/av/audio/format.pyx
@@ -1,3 +1,4 @@
+# cython: c_string_type=unicode, c_string_encoding=ascii
 import sys
 
 
--- a/av/audio/layout.pyx
+++ b/av/audio/layout.pyx
@@ -1,3 +1,4 @@
+# cython: c_string_type=unicode, c_string_encoding=ascii
 cimport libav as lib
 
 
--- a/av/codec/codec.pyx
+++ b/av/codec/codec.pyx
@@ -351,7 +351,7 @@ codec_descriptor = wrap_avclass(lib.avcodec_get_class())
 def dump_codecs():
     """Print information about availible codecs."""
 
-    print '''Codecs:
+    print('''Codecs:
  D..... = Decoding supported
  .E.... = Encoding supported
  ..V... = Video codec
@@ -360,7 +360,7 @@ def dump_codecs():
  ...I.. = Intra frame-only codec
  ....L. = Lossy compression
  .....S = Lossless compression
- ------'''
+ ------''')
 
     for name in sorted(codecs_available):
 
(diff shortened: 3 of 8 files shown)


mikeboers:
There is a Cython 3 now? Cool. 😛 

(Aside, I wish there was a nice summary of major changes in Cython versions. I feel like I never really know what is important in the changelog.)

A few thoughts:

- There are a lot of failing test runs.
- I would like to see the Cython encoding directives centralized [like this](https://github.com/PyAV-Org/PyAV/blob/b43a7864abed7be9a97a58ed27e66067c7ba5081/setup.py#L417-L421).
- Surely [this](https://github.com/PyAV-Org/PyAV/pull/712/commits/d36c8df02e37097b5bf736fc2b934e57264cff04#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R29-R31) can be done by passing something to the `cythonize` function.
- Can you offer any further justification to the change to relative imports?

KOLANICH:
>Can you offer any further justification to the change to relative imports?

* Relative imports allow easier refactoring of the package. DRY principle.
* And they make imports a bir shorter and easier to read.

KOLANICH:
>https://github.com/PyAV-Org/PyAV/blob/b43a7864abed7be9a97a58ed27e66067c7ba5081/setup.py#L421

It seems it doesn't work, cyyhon complains to me about language_level not being set

KOLANICH:
>Surely this can be done by passing something to the cythonize function

I tried hard. It doesn't work like that. I have spent a lot of time by trying to make that work, even thought about a bug in cython and tried to workaround that by moving includes into src dir ... It worked, so since I already had to use relative import to import includes, I have replwced al, the imports with relative. Then I thought that since cython mimics python, that its imports should work like python ones. So I moved the includes back and just added them into path. That was a brilliant guess. Though I haven't removed the relative paths, since IMHO these changes are needed anyway.

KOLANICH:
>Aside, I wish there was a nice summary of major changes in Cython versions.

Cython 3 is not released, it is a version number used by the version of cython installed from its git repo. But it works, and I prefer use bleeding edge versions of software and libs.

jlaine:
> > Can you offer any further justification to the change to relative imports?
> 
>     * Relative imports allow easier refactoring of the package. DRY principle.
> 
>     * And they make imports a bir shorter and easier to read.

I have to admit I'm not sold on the relative imports, I fail to see what we're gaining, and for sure it makes grepping harder..

jlaine:
Do you plan to finish this?

KOLANICH:
I have no time for working on this PR currently. I have implemented this PR long ago, I don't remember exactly if it was working that time. Everyone is welcome to use this PR as a base for an own PR.

KOLANICH:
https://github.com/cython/cython/releases/tag/3.0.0

PR #713, opened 2020-10-14
.gitignored generated *.c files

KOLANICH:


mikeboers:
All the `*.c` files should be landing in the `/src` directory which is already ignored. If this is related to your use of Cython3 it should be addressed there.

PR #714, opened 2020-10-15
[build] upgrade cibuildwheel to 1.6.3

jlaine:


jlaine:
@mikeboers Hi! This PR resolves two issues:

- the trickiest: we had to pin `cibuildwheel` to 1.4.2 due to changes in the way `CIBW_ENVIRONMENT` is parsed in more recent version. We now *need* this upgrade, as otherwise we get build failures on OSX.

- the easiest: isort now also applies to *.pxd files, so the imports needed a bit of shuffling.

The problem I now have is that the testsuite executed on the wheels fails the "unicode filename" test on Python 3.7. What I don't understand is that the same test passes, see tests/py-3.7* on Windows. Unless you understand what's going on I'm tempted to skip the test entirely on Windows, so that we get CI back to a sane state.

mikeboers:
Sorry @jlaine. Skipping the test is likely the best thing at the moment. 😊

PR #718, opened 2020-11-10
Implement custom io_open/io_close support in container

philipnbbc:
This PR adds an `io_open` option to `av.open()` which allows a client to provide a custom I/O. The `io_open` function is called with parameters `url`, `flags`, and `options` when a new file needs to be opened.

The existing file I/O was factored out into the `PyIOFile` class, with the addition of the `fclose` method. This class is used for the main file as well as for any files opened by the container.

Custom I/O has been used with the FFmpeg DASH encoder and could be used with containers that uses the `io_open` and `io_close` callbacks in the AVFormatContext. The test shows how it can be used with the DASH encoder by using a CUSTOM_IO_PROTOCOL prefix to direct the DASH encoder to use the custom I/O.


jlaine:
I like the fact that the Python I/O code gets moved to a single place, I think this deserves to be a PR in it own right.

What I don't understand is the purpose of the new `io_open` argument. Can't this be achieved by writing your own file-like object and passing it directly to av.open ?

Also:

- I don't understand the `open_files` dict. Why would there be multiple open files per container?
- The test suite emits a new warning: `Cannot use rename on non file protocol, this may lead to races and temporary partial files`

jlaine:
Hi @philipnbbc I really like the look of this PR, but I'm getting a segfault when trying to use custom I/O to read back the written files. I tried adding the following to `test_writing_to_custom_io`:

```python
        # Check contents.
        with av.open(
            CUSTOM_IO_PROTOCOL + CUSTOM_IO_FILENAME, "r", io_open=wrapped_custom_io
        ) as container:
            assert_rgb_rotate(self, container)
```

Any idea what's going on?

jlaine:
I have merged a first block of code : the factoring out into the `PyIOFile` class. For the rest of the code, I'll wait for your feedback on the crash I encountered when reading.

philipnbbc:
Thanks for taking a look at this PR. I'll take a look at the warnings and segmentation fault hopefully this week.

The custom I/O is intended to support multi-file containers such as DASH where secondary files (e.g. DASH segment files) are opened by FFmpeg (using the custom I/O API). It wasn't intended to be used to open the primary file (e.g. DASH manifest) because as you say, the user can pass in a file-like object. My guess is that the segmentation fault happens in your example because the container doesn't exist at the point the file is opened.

I remember that warning but don't remember whether it was there before or that something has changed in FFmpeg

philipnbbc:
Regarding the error "Cannot use rename on non file protocol, this may lead to races and temporary partial files":

Support for non-file protocols was added to DASH in commit https://github.com/FFmpeg/FFmpeg/commit/5fe2b437023f46394dfd4a4c991aa4a3ac3d8e72. The commit includes the error log message at https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/dashenc.c#L1145 that is emitted when a non-file protocol is used. The commit message also states that the code logic was copied from the HLS format.

However, hlsenc.c includes a HLS_TEMP_FILE flag to indicate that temporary files should be used, https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/hlsenc.c#L1572. In that case it makes sense to log an error if the flag is set and a non-file protocol is used. I think in the DASH case it should not emit an error message because there is no flag to enable use of temporary files / renaming.

philipnbbc:
> * I don't understand the `open_files` dict. Why would there be multiple open files per container?

The custom I/O doesn't specify that formats that read or write multiple files (like DASH) only have one file open at the same time. It's possible that the DASH implementation does close files before opening new ones but I didn't want to make that assumption to avoid breakages in future.

philipnbbc:
> Hi @philipnbbc I really like the look of this PR, but I'm getting a segfault when trying to use custom I/O to read back the written files. I tried adding the following to `test_writing_to_custom_io`:
> 
> ```python
>         # Check contents.
>         with av.open(
>             CUSTOM_IO_PROTOCOL + CUSTOM_IO_FILENAME, "r", io_open=wrapped_custom_io
>         ) as container:
>             assert_rgb_rotate(self, container)
> ```
> 
> Any idea what's going on?

The segmentation fault was caused by file open failure block closing the custom I/O when it shouldn't and that was fixed by adding the AVFMT_FLAG_CUSTOM_IO flag - fixed in https://github.com/PyAV-Org/PyAV/pull/718/commits/a4c472bc33202c0e6c329f0024417e21b44204ff.

The next issue was that the DASH demuxer wasn't enabled in the ffmpeg build used in the tests - fixed in https://github.com/PyAV-Org/PyAV/pull/718/commits/ce2d70054d701b5aeb40415d5d827da584b33a94.

The DASH demuxer doesn't support segment files relative to the manifest that is in another directory. That required the content check, added in https://github.com/PyAV-Org/PyAV/pull/718/commits/4ebd45486142da3bd6bb2cffa7d62bad55842ce5, to run the test in the sandbox directory.

The important thing to note here is that the DASH demuxer does not support custom I/O and so using `io_open=wrapped_custom_io` will only result in the manifest being opened using the custom I/O. Also, the CUSTOM_IO_PROTOCOL isn't required for reading and will result in an error anyway because the DASH demuxer only supports the file and http protocols.

scripts/build-deps
```diff
@@ -44,6 +44,8 @@ echo ./configure
     --enable-debug=3 \
     --enable-gpl \
     --enable-libx264 \
+    --enable-libxml2 \
```
jlaine, line 47: Why these changes?

tests/test_encode.py
```diff
@@ -77,7 +82,12 @@ def assert_rgb_rotate(self, input_):
         stream.average_rate, 24
     )  # Only because we constructed is precisely.
     self.assertEqual(stream.rate, Fraction(24, 1))
-    self.assertEqual(stream.time_base * stream.duration, 2)
+    if stream.duration is not None:
```
jlaine, line 85: I'm not keen of these changes, I don't like tests which depend on the output. How about adding an explicit `is_dash` argument to assert_rgb_rotate?

tests/common.py
```diff
@@ -72,6 +73,17 @@ def sandboxed(*args, **kwargs):
     return path
 
 
+# Context manager for running a test in a directory, e.g. path is the sandbox
+@contextmanager
+def run_in_directory(path):
```
jlaine, line 78: Or this could be a decorator?

scripts/build-deps
philipnbbc: I've simplified it in https://github.com/PyAV-Org/PyAV/pull/718/commits/cc5b261e80122e46853609bb01f718becf6da26e to just enable libxml2. The DASH demuxer, which is used in the tests, is not enabled if libxml2 is not enabled; libxml2 is used to read the DASH manifest XML file.

tests/test_encode.py
philipnbbc: Changed in https://github.com/PyAV-Org/PyAV/pull/718/commits/d1a9c422924d613c2bc9996828b12da46bf6e273. I noted a build failure on FFmpeg 4.1 and have added a fix for that too.

tests/common.py
philipnbbc: Changed in  https://github.com/PyAV-Org/PyAV/pull/718/commits/90a3479a7d57676e8f91b24b50803536097fdacc

tests/test_encode.py
jlaine: Looks good!

av/container/core.pyx
```diff
@@ -251,7 +333,7 @@ cdef class Container(object):
 def open(file, mode=None, format=None, options=None,
          container_options=None, stream_options=None,
          metadata_encoding='utf-8', metadata_errors='strict',
-         buffer_size=32768, timeout=None):
+         buffer_size=32768, timeout=None, io_open=None):
```
jlaine, line 336: This `io_open` argument needs a docstring explaining what the expected signature is, and how to use it.

scripts/build-deps
jlaine: Would a test which writes a sequence of PNG files also make sense?

philipnbbc: I think that would be worth doing (added in https://github.com/PyAV-Org/PyAV/pull/718/commits/d4b23725cb18303bdf809730156889d2c452e92b) and it actually exposed a bug that the DASH muxer hadn't (fixed in https://github.com/PyAV-Org/PyAV/pull/718/commits/f77e71464a259a72fb199fa76dfe7fe3186832bd).

tests/test_encode.py
```diff
@@ -67,9 +67,12 @@ def assert_rgb_rotate(self, input_, is_dash=False):
 
     # Now inspect it a little.
     self.assertEqual(len(input_.streams), 1)
-    if is_dash:
-        # The "title" metadata is named "Title" in the DASH format
-        self.assertEqual(input_.metadata.get("Title"), "container", input_.metadata)
+    if is_dash and input_.metadata.get("title") is None:
```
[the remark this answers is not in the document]
philipnbbc: That's useful to know. I've updated that commit to use it, https://github.com/PyAV-Org/PyAV/pull/718/commits/67b9ce05f8ca125c62e7235e6c7d094b8dd8cc30

av/container/core.pyx
philipnbbc: Added in https://github.com/PyAV-Org/PyAV/pull/718/commits/d84cc95b17d8dae6ac2b3329851d00fc7cda0d76

philipnbbc:
@jlaine I've made the changes you requested. I assume the commits should be squashed before merging, once the tests have succeeded and you don't have anymore comments.

jlaine:
Amazing work @philipnbbc, thanks for your perseverance! The new docstrings are very clear, and you've done a stellar job on the unit tests. Merging!

PR #719, opened 2020-11-10
fix docs: loglevel QUIET isn't available

pklaus:
The documentation lists `QUIET` as a possible loglevel since 64e80ee.
In that same commit, QUIET was marked as `not actually a level`, however.
Later it was fully removed as a loglevel in 2ee9684.

This pull request / commit fixes the documentation in that regard.

### DIFF
--- a/av/logging.pyx
+++ b/av/logging.pyx
@@ -106,7 +106,7 @@ def set_level(int level):
 
     Sets logging threshold when converting from FFmpeg's logging system
     to Python's. It is recommended to use the constants availible in this
-    module to set the level: ``QUIET``, ``PANIC``, ``FATAL``, ``ERROR``,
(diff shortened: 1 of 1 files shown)


av/logging.pyx
```diff
@@ -106,7 +106,7 @@ def set_level(int level):
 
     Sets logging threshold when converting from FFmpeg's logging system
     to Python's. It is recommended to use the constants availible in this
```
jlaine, line 108: While we are fixing this section could we correct "availible" => "available" please?

pklaus:
Great, thanks!

jlaine:
Thanks for your PR and your patience. PyAV is kind of time-intensive to maintain, and both @mikeboers and I have limited time at the moment..

Issue #720, opened 2020-11-11
pkg-config returned flags we don't understand: -pthread -pthread


osynge:
## Overview
When installing PyAV with the following command:

    pip install av --no-binary av 

The log shows the error:

    pkg-config returned flags we don't understand: -pthread -pthread

This is due to incorrect passing of the result of `pkg-config` which outputs lines like:

    root@cd3ee8976f50:/# pkg-config --cflags --libs  libavformat
    -lavformat -lm -lbz2 -lz -lavcodec -pthread -lm -llzma -lcrystalhd -lz -lva -ldl -lswresample -lm -lavutil -pthread -lva-drm -lva -lva-x11 -lva -lvdpau -lX11 -lm -lva -lXv -lX11 -lXext -ldl
    root@cd3ee8976f50:/# pkg-config --cflags --libs  libavcodec
    -lavcodec -pthread -lm -llzma -lcrystalhd -lz -lva -ldl -lswresample -lm -lavutil -pthread -lva-drm -lva -lva-x11 -lva -lvdpau -lX11 -lm -lva -lXv -lX11 -lXext -ldl
    root@cd3ee8976f50:/# pkg-config --cflags --libs  libavdevice
    -lavdevice -lm -lxcb -lxcb-shm -lxcb-shape -lxcb-xfixes -lasound -lSDL2 -lsndio -lXv -lX11 -lXext -lavfilter -pthread -lm -lnppig -lnppicc -lnppc -lnppidei -lva -ldl -lswscale -lm -lavformat -lm -lbz2 -lz -lavcodec -pthread -lm -llzma -lcrystalhd -lz -lva -ldl -lswresample -lm -lavutil -pthread -lva-drm -lva -lva-x11 -lva -lvdpau -lX11 -lm -lva -lXv -lX11 -lXext -ldl
    root@cd3ee8976f50:/# pkg-config --cflags --libs  libavutil
    -lavutil -pthread -lva-drm -lva -lva-x11 -lva -lvdpau -lX11 -lm -lva -lXv -lX11 -lXext -ldl
    root@cd3ee8976f50:/# pkg-config --cflags --libs  libavfilter
[...]

PR #721, opened 2020-11-11
pkg-config returned flags we don't understand: -pthread -pthread

osynge:
When installing with `pip install av --no-binary av` we get the
error shown above, this change allows setup.py to ignore -pthread

Resolves [bug 720](https://github.com/PyAV-Org/PyAV/issues/720)


jlaine:
To the best of my knowledge this error only crops up when building against a statically linked FFmpeg, which is not something we support.

PR #725, opened 2020-12-09
example: encoding with custom PTS

crackwitz:
I didn't notice an encoding example that uses PTSes so I thought I'd add one.

I hope I got this right. I realize ffmpeg's notions of time are "implementation-defined", to put it mildly, and PyAV can't hide that in any good way.

 - setting `stream.codec_context.time_base`
 - setting `frame.pts` calculated from `stream.codec_context.time_base`

### DIFF
--- a/examples/numpy/generate_video_with_pts.py
+++ b/examples/numpy/generate_video_with_pts.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+
+from fractions import Fraction
+import colorsys
+
+import numpy as np
+
+import av
+
+
+(width, height) = (640, 360)
+total_frames = 20
+fps = 30
+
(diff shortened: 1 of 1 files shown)


jlaine:
This is very neat, thanks for taking the time to put this PR together. Do you think we could have an entry in `docs/cookbook/basics.rst` which references this example please?

gstrat88:
Regarding this example, is it possible to somehow reset timestamps when we are dealing with a copy stream/remux situation?

https://pyav.org/docs/develop/cookbook/basics.html#remuxing

crackwitz:
basics.rst... if I understand correctly, you're asking for a pull request to add a section there. I'll try to remember to do that.

jlaine:
Thanks that would be great, as currently there is a low chance of people discovering your example!

PR #729, opened 2020-12-26
[tests] do not hardcode errno values

pinotree:
Do not assume that `1 == ENOENT` and `2 == EPERM` (as `OSError`); rather use the `errno` module to get/use their values.

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,3 +1,4 @@
+import errno
 import traceback
 
 import av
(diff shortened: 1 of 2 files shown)


jlaine:
For my information, on what platforms is this not true?

pinotree:
> For my information, on what platforms is this not true?

This was on GNU/Hurd. It does not matter though, as POSIX does not define the actual values of errno constants, so relying on them is always not correct.

crackwitz:
> `1 == ENOENT` and `2 == EPERM`

however pull request replaces 1 with EPERM and 2 with ENOENT... perhaps give this another look and maybe try provoking these errors for a quick manual test

pinotree:
> comment isn't.

Hmm, I do not understand what you mean, can you please explain it?

crackwitz:
https://www.man7.org/linux/man-pages/man3/errno.3.html
> EPERM 1 Operation not permitted
> ENOENT 2 No such file or directory

pull request matches. previous comment switched them up. that's all I'm saying, if anyone decided to believe the comment over the code. just ignore what I said. maybe I'm just confusing everyone.

jlaine:
@crackwitz nice catch regarding the commit message, I fixed it an merged the PR. Thanks @pinotree !

Issue #730, opened 2021-01-01
Can't use stream added from template for anything other than remuxing


alexhutnik:
## Overview

When I create a new Container, and add a new stream to it using a template, I can only remux packets. I cannot decode and then re-encode packets.

## Expected behavior

I would expect to be able to roundtrip a packet like `decode packet` -> `frame` -> `encode frame` -> `packet`.

## Actual behavior

When trying to do this roundtrip, I am presented with an error.

Traceback:
```
File "av/stream.pyx", line 152, in av.stream.Stream.encode
File "av/codec/context.pyx", line 476, in av.codec.context.CodecContext.encode
File "av/codec/context.pyx", line 391, in av.codec.context.CodecContext._send_frame_and_recv
File "av/error.pyx", line 336, in av.error.err_check
av.error.ValueError: [Errno 22] Invalid argument
```

## Investigation

I've dug about as far in to the pyav source as I can, but I'm not very familiar with Cython

## Reproduction

```
input_container = av.open('in.mp4', 'r')
input_stream = input_container.streams.get(video=0)[0]

output_container = av.open('out.mp4', 'w')
output_stream = output_container.add_stream(template=input_stream)

for frame in input_container.decode(input_stream):
    # encode and mux
    for packet in output_stream.encode(frame):
        packet.pts = frame.pts
        packet.time_base = frame.time_base
        # packet.stream = output_stream
        output_container.mux(packet)

output_container.close()
```

## Versions

- OS: Ubuntu 18.04
- PyAV runtime:
```
PyAV v8.0.2
git origin: git@github.com:PyAV-Org/PyAV
git commit: v8.0.2
library configuration: --disable-doc --disable-static --enable-fontconfig --enable-gmp --enable-gnutls --enable-gpl --enable-libaom --enable-libass --enable-libbluray --enable-libdav1d --enable-libfreetype --enable-libmp3lame --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopenjpeg --enable-libopus --enable-libspeex --enable-libtheora --enable-libtwolame --enable-libvorbis --enable-libwavpack --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libxvid --enable-lzma --enable-shared --enable-version3 --enable-zlib --prefix=/tmp/vendor
library license: GPL version 3 or later
libavcodec     58. 54.100
libavdevice    58.  8.100
libavfilter     7. 57.100
libavformat    58. 29.100
libavutil      56. 31.100
[...]

Issue #507, opened 2019-04-17
Encoding frames error: Errno 22 Invalid Argument


flewellyn:
I am attempting to create a "preview" of a video, using key frames.  Here is the logic:

```invid = av.open('001.mov')
instream = invid.streams.video[0]

outvid = av.open('001_prev.mov')
outstream = outvid.add_stream(template=instream)

outstream.time_base = instream.time_base

instream.codec_context.skip_frame = 'NONKEY'

for frame in invid.decode(instream):
    packet = outstream.encode(frame)
    outvid.mux(packet)

outvid.close()
```

According to the docs, this should work, but the outtream.encode() call fails with Errno 22.  I have no idea why.  The frame itself is correct: when I use frame.to_image() and save a JPEG, it works just fine.

jlaine:
The `Stream.encode` method returns packetS, not a single packet.

Try:

```python

# transcode
for frame in invid.decode(instream):
    for packet in outstream.encode(frame):
        outvid.mux(packet)

# flush encoder
for packet in outstream.encode(None):
    outvid.mux(packet)
```

flewellyn:
Hmm, no, that doesn't work either, I'm afraid.  I get the same error.

jlaine:
Can you provide us with the file you are using for tests?

**Important** : make sure you are the copyright holder of the file!

iyah4888:
I have the exactly same issue as well on pyav-6.2.0.

I'm transcoding, and input frame is decoded as the following object:
<av.VideoFrame #4, pts=2048 yuv420p 1280x720 at 0x7fee6368eee8>
and the output stream is 
<av.VideoStream #0 h264, yuv420p 1280x720 at 0x7fee63c30108>

`
datapath = 'ewkFvB89_AE.00278244_00281414.mp4'
input_ = av.open(datapath)
in_stream_v = input_.streams.video[0]
in_stream_v.thread_type = 'AUTO'

output = av.open('test.mp4', 'w')
out_stream_v = output.add_stream(template=in_stream_v)


for packet in input_.demux(in_stream_v):
    if packet.dts is None:
        continue
        
    print(packet)
    
    for frame in packet.decode():
        for packet_out in out_stream_v.encode(frame):
            output.mux(packet_out)

# Flush stream
# Close the files
`

jlaine:
Again: files or it didn't happen.

iyah4888:
[...]

PR #732, opened 2021-01-06
Stream copy from template

alexhutnik:
Addresses issue #730 and #507. When adding a stream via a template, look up the encoder via the template codec's id, consistent with Transcoding.c in the ffmpeg source: https://github.com/FFmpeg/FFmpeg/blob/66deab3a2609aa9462709c82be5d4efbb6af2a08/doc/examples/transcoding.c#L149


alexhutnik:
Somehow this broke regular remuxing and I didn't catch it before the PR. Not ready to merge.

alexhutnik:
There seems to be something that causes a packet to be owned by the AVCodecContext that demux'd it. When you try to mux it using a different AVCodecContext, it silently fails to write anything.  So you can either copy the context as originally done, or you can construct a new one, which requires one to decode / re-encode, even if the packet is never modified.  I've added a `remux` argument to the add_stream method to pick which behavior to go with.  I'll leave this PR open for now.

include/libavcodec/avcodec.pxd
```diff
@@ -116,6 +116,92 @@ cdef extern from "libavcodec/avcodec.h" nogil:
         AVDISCARD_NONKEY
         AVDISCARD_ALL
 
+    cdef enum AVChromaLocation:
```
jlaine, line 119: Why all the new enums?

alexhutnik: Sorry, just saw this. My GitHub notifications are a mess. 

I had to add them because of the declared types in this method: https://github.com/PyAV-Org/PyAV/blob/1213f02dc28a14aa803753a8b57f9eae9e07604a/include/libavcodec/avcodec.pxd#L533

alexhutnik:
@jlaine Overall I wonder if this PR is still necessary given the work being done to deprecate avcodec_copy_context usage. I'm not sure if the core issue still stands: Creating an encoding context whose configuration is copied from a decoder is not setting the correct context flag. This causes the encoding context to actually think it is a decoder because the context is copied too literally. Then, when you tell it to encode a packet, it can't.

If I can get to it, I'll check this out in the latest PyAV with ffmpeg 5 and see if I can recommend an approach.

HaujetZhao:
I met the same problem with you. 

I want to create an output video stream that first remuxes packets from an input video and then encodes frames from another source. The code might look like this:

```
import av

input_ = av.open('input.mp4')
output = av.open('output.mp4', "w")

in_stream = input_.streams.video[0]
out_stream = output.add_stream(codec_name=in_stream.codec_context.codec.name)

# First, the output stream remuxes packets from the input video
for packet in input_.demux(in_stream):
    if packet.dts is None:
        continue
    packet.stream = out_stream
    output.mux(packet)

# Get some av.Frames from another source (e.g., images, generated frames, etc.)
frames = []
# ... Do some processing to fill the frames list with av.Frame objects

# Then, the output stream encodes frames from the other source
for f in frames:
    for packet in out_stream.encode(f):
        output.mux(packet)

input_.close()
output.close()
```
Unfortunately, this simple code doesn't work as expected. The output stream doesn't successfully mux the packets from the input stream in the first step.

I know I can use out_stream = output.add_stream(template=in_stream) to successfully remux the packets, but in that case, the output stream will only get a decoder codec, which can't be used to encode the frames in the second step. 

The created output stream can only either remux or encode. Only capable of doing one thing. So If I want to append some piece of frames or clips, I must re-encode the whole original video frames. 

Of course, a work around is to use FFmpeg [concat](https://trac.ffmpeg.org/wiki/Concatenate) , but the concat can't append frames or clips that lively generated. 

So my question is, is it possible to create an encoder that can mux packets demuxed from a decoder, Is it a problem that has no solution at its root, or it can be solved by making some adjustment on the output_stream.context as you are doing now?

WyattBlue:
I"m not sure if these changes even make sense anymore with ffmpeg 6. Besides, the changes needs to be based on `main`

Issue #741, opened 2021-02-12
Allow vtag to be set


iBobbyTS:
## Overview
In FFmpeg we're able too set vtag for a stream, for example hvc1 is a tag of HEVC, so that we can play with QuickTime Player on a Mac (QuickTime only plays HEVC streams with the vtag hvc1, but the default vtag of libx265 is hev1). 

## Existing FFmpeg API
I'm not familiar with the C API of FFmpeg, I can only tell the command line option that set the vtag. 
`ffmpeg -i input.mp4 -c:v libx265 -vtag hvc1 output.mov`
OR
`ffmpeg -i input.mp4 -c:v libx265 -tag:v hvc1 output.mov`

## Expected PyAV API
Like what setting pix_fmt is like. 
https://pyav.org/docs/stable/cookbook/numpy.html#generating-video

Example:
```
container = av.open('test.mov', mode='w')
stream = container.add_stream('libx265', rate=30)
stream.tag = 'hvc1'
```

## Versions

- OS: {{ e.g. macOS 11.2.1 }}
- PyAV runtime:
```
PyAV v8.0.3
git origin: git@github.com:PyAV-Org/PyAV
git commit: v8.0.3
library configuration: --disable-doc --disable-static --enable-fontconfig --enable-gmp --enable-gnutls --enable-gpl --enable-libaom --enable-libass --enable-libbluray --enable-libdav1d --enable-libfreetype --enable-libmp3lame --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopenjpeg --enable-libopus --enable-libspeex --enable-libtheora --enable-libtwolame --enable-libvorbis --enable-libwavpack --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libxvid --enable-lzma --enable-shared --enable-version3 --enable-zlib --prefix=/tmp/vendor
library license: GPL version 3 or later
libavcodec     58. 91.100
libavdevice    58. 10.100
libavfilter     7. 85.100
libavformat    58. 45.100
libavutil      56. 51.100
libswresample   3.  7.100
libswscale      5.  7.100
```
- PyAV build:
`Installed from pip`
- FFmpeg:
`Installed from pip`

mikelei8291:
[...]

PR #748, opened 2021-02-25
Expose codec_tag in codec_context

uvjustin:
This PR exposes codec_tag as a property of codec_context as discussed here: https://github.com/PyAV-Org/PyAV/issues/741


jlaine:
If you could add some unit tests for this, this looks good and mergeable.

uvjustin:
Added a simple test for getting/setting the tag and encoding with the tag set. I didn't see an easy way to assert that the codec actually used the tag, so that part is untested.

av/utils.pxd
```diff
@@ -9,5 +9,7 @@ cdef dict_to_avdict(lib.AVDictionary **dst, dict src, str encoding, str errors)
 cdef object avrational_to_fraction(const lib.AVRational *input)
 cdef object to_avrational(object value, lib.AVRational *input)
 
+cdef object name_to_codec_tag(object name)
```
jlaine, line 12: Some questions about types:

- I see codec_tag is declared as `unsigned int` (so length can depend on the platform), not `uint32_t`
- how come name_to_codec_tag returns an object and not an `unsigned int`?

jlaine:
Thanks so much for adding the tests, that's awesome!

@mikeboers do you have any input on the comments I made about types?

av/utils.pxd
uvjustin: Good points. I'll change the codec_tag type to `unsigned int` (although I'm not sure how the tags are supposed to work when `unsigned int` is 2 bytes).
I'll make the other changes as well. I think `str` is a more readable/user friendly type to expose as long as we account for errors.
This is my first time touching Cython, so please forgive any obvious problems.

jlaine: There's nothing obvious about these problems, I didn't spot them at first! Do you have any examples of non-ascii tags used by codecs?

jlaine: I had a look at how the FFmpeg takes tags from the command line, and I *think* it's manipulating them as uint32t:

https://github.com/FFmpeg/FFmpeg/blob/master/fftools/ffmpeg_opt.c#L821-L827

uvjustin: Moved it to draft as I'm playing around with using python built-ins instead of adding extra utility functions.

uvjustin: Changed it to use the utility functions. The getter shouldn't have any problems since we're converting from the c integer type to a more flexible python int type. Will the setter have issues if we try to set an overflowing integer into a 2 byte unsigned int?

uvjustin: I guess it will just raise an OverflowError, but we probably don't have to handle it

av/codec/context.pyx
```diff
@@ -540,6 +540,18 @@ cdef class CodecContext(object):
         def __set__(self, value):
             to_avrational(value, &self.ptr.time_base)
 
+    property codec_tag:
+        def __get__(self):
+            return self.ptr.codec_tag.to_bytes(4, byteorder="little", signed=False).decode(
```
jlaine, line 545: I would say stick to ASCII, i don't see anything suggesting this should allow itf8 in a fixed 4 byte budhet

uvjustin: Changed to ASCII

jlaine:
Crap I messed up my push.

uvjustin:
> 
> 
> Crap I messed up my push, and I probably screwed up your `main` branch in your repo, I'm really sorry.
> 
> Anyway, the PR lives on as #750

No problem!

PR #749, opened 2021-02-25
Fix args order in Frame.__repr__

MarSoft:
This code is probably unused, but the bug is still there


jlaine:
Hm, both AudioFrame and VideoFrame do redefine their `__repr__` methods so I'm not sure it's used. Still you're right, this is wrong.

PR #750, opened 2021-02-26
Expose codec_context.codec_tag (fixes: #741)

jlaine:
This is just #748 with an additional test

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -540,6 +540,18 @@ cdef class CodecContext(object):
         def __set__(self, value):
             to_avrational(value, &self.ptr.time_base)
 
+    property codec_tag:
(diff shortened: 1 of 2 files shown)


uvjustin:
The other problem is that the current get/set test only really tests that the get and the set are compatible with each other - they don't really test the representation is stored into the codec tag correctly. A related issue here is that 'divx' and 'xvid' are both valid tags so a problem with the endianness might be hard to spot.
However, I tested it locally with `libx265` and the `hvc1` tag and the output confirmed that the tag was working properly.

jlaine:
I agree, we're only testing the consistency of our getter/setter. Do you know if any files in the FATE test suite might allow us to get the tag from an existing file?

PR #752, opened 2021-02-26
Update test to check codec tag against known FATE suite sample

uvjustin:
Updated the test to check against a FATE suite sample as suggested in #750


jlaine:
Fantastic, thanks for this!

PR #759, opened 2021-03-05
Add rgb48 and rgba64 pixel formats

apetrynet:
Hi!

I was testing out PyAV for a dailies/deliveries tool I'm writing and I noticed there was no support for uint16 data type in the numpy integration. 
With the added pixel types this PR I'm able to create ProRes video files based on 16bit images and maintain image quality.

I only added support for the little endian formats as I know too little on the matter and (_quite selfishly_) I don't know if I need support for big endian..

### DIFF
--- a/av/video/frame.pyx
+++ b/av/video/frame.pyx
@@ -49,7 +49,7 @@ cdef copy_array_to_plane(array, VideoPlane plane, unsigned int bytes_per_pixel):
         o_pos += o_stride
 
 
-cdef useful_array(VideoPlane plane, unsigned int bytes_per_pixel=1):
+cdef useful_array(VideoPlane plane, unsigned int bytes_per_pixel=1, str dtype='uint8'):
     """
     Return the useful part of the VideoPlane as a single dimensional array.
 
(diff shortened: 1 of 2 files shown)


jlaine:
Hi! This looks cool, thanks for putting it together. One thing leaves me a bit dubious: since we are talking about multi-byte formats, how do we know numpy's arrays are storing data in the expected endianness? Wouldn't this be platform-dependent?

apetrynet:
Good point!
I can perhaps check the byteorder of the dtype and make sure it matches what the end user is asking for. I'll do some tests. Who knows, maybe I'll end up adding support for the rgb48be and rgba64be pixel formats after all :)

Thanks for the feedback!

L3wisXI (approved):

jlaine:
Hi! Having recently fixed some endianness-related bugs in AudioFrame, I'm a little dubious about how this code works. I haven't forgot your PR though!

apetrynet:
Hi!
Sorry about the late reply.
I'm not sure I understand. Are you saying my endianess-solution is a bit dodgy or that you don't quite get what I've done? 
It's been a while since I last looked at this so I'll need to refresh my memory in any case :)

jlaine:
Hi @apetrynet I have added an alternative implementation in PR #939 

The key difference here is that the ndarray are always in the host's native system, which should help reduce confusion. Any thoughts on this?

jlaine:
I'm closing this in favor of #939

apetrynet:
Apologize again for late reply. I've been quite busy and I needed time to put my head into this again. 
Anyway, glad you managed to solve this in your PR.

PR #765, opened 2021-03-27
Better time_base support with filters

fvollmer:
PyAV frames didn't have a `time_base` after they were passed through a filter and `add_buffer` also ignored the `time_base`.

The following example deinterlaces a 30fps video.
```python
import av.filter
import errno

input_container = av.open(format='lavfi', file='color=c=pink:duration=5:r=30')
input_video_stream = input_container.streams.video[0]

graph = av.filter.Graph()
buffer = graph.add_buffer(template=input_video_stream)
bwdif = graph.add("bwdif", "send_field:tff:all")
buffersink = graph.add("buffersink")
buffer.link_to(bwdif)
bwdif.link_to(buffersink)
graph.configure()

for frame in input_container.decode():
    print(f"old time_base: {frame.time_base}, old pts: {frame.pts}")
    graph.push(frame)
    while True:
        try:
            filtered_frame = graph.pull()
        except av.utils.AVError as e:
            if e.errno != errno.EAGAIN:
                raise
            break

        print(f"filtered time_base: {filtered_frame.time_base}, filtered pts: {filtered_frame.pts}")
```
With these patches the time_base is handled correctly
```
old time_base: 1/30, old pts: 0
old time_base: 1/30, old pts: 1
filtered time_base: 1/60, filtered pts: 0
filtered time_base: 1/60, filtered pts: 1
old time_base: 1/30, old pts: 2
filtered time_base: 1/60, filtered pts: 2
filtered time_base: 1/60, filtered pts: 3
old time_base: 1/30, old pts: 3
filtered time_base: 1/60, filtered pts: 4
filtered time_base: 1/60, filtered pts: 5
...
```
instead of
```
old time_base: 1/30, old pts: 0
old time_base: 1/30, old pts: 1
filtered time_base: None, filtered pts: 0
filtered time_base: None, filtered pts: 1
old time_base: 1/30, old pts: 2
filtered time_base: None, filtered pts: 2
filtered time_base: None, filtered pts: 3
old time_base: 1/30, old pts: 3
filtered time_base: None, filtered pts: 4
filtered time_base: None, filtered pts: 5
```

PR #766, opened 2021-03-28
Rework resampler and fifo to match ffmpeg behavior

### LINKED ISSUE #761: pts trouble with audio from dshow
I'm recording audio from dshow and there seems to be a problem with the pts. ffmpeg isn't complaining about these sources, but I also think this might not be a bug in pyav. I tried several input devices and it always creates some error like:
```
Encoder did not produce proper pts, making some up.
Traceback (most recent call last):
  File "C:/Users/User/PycharmProjects/project/audio.py", line 43, in <module>
    output_container.mux(output_audio_stream.encode(frame))
  File "av\stream.pyx", line 155, in av.stream.Stream.encode
  File "av\codec\context.pyx", line 466, in av.codec.context.CodecContext.encode
  File "av\audio\codeccontext.pyx", line 40, in av.audio.codeccontext.AudioCodecContext._prepare_frames_for_encode
  File "av\audio\resampler.pyx", line 122, in av.audio.resampler.AudioResampler.resample
ValueError: Input frame pts 980000 != expected 1000000; fix or set to None.
```
I can obviously just set the pts to `None` and it will just make the some up. This seems to be deprecated and results in the following warning
```
Timestamps are unset in a packet for stream 0. This is deprecated and will stop working in the future. Fix your code to set the timestamps properly
Encoder did not produce proper pts, making some up.
```

To debug the problem I wrote the following code, which prints pts, time base, sample rate, samples etc. and allows to see the what exactly is going on.
```python
import av

# doesn't work with dshow (tried different devices, and also specifying the sample rate)
input_container = av.open(format='dshow',
                          file='audio=Eingang (High Definition Audio Device)',
                          #file='audio=Mikrofon (USB2.0 MIC)',
                          options={"audio_buffer_size": "100",
                                   #"sample_rate": "44100",
                                   },
                          )

# # works
# input_container = av.open(format='lavfi', file='sine=frequency=1000:duration=5',
#                           options={"sample_rate": "44100"})

input_audio_stream = input_container.streams.audio[0]

output_container = av.open('test.mkv', mode='w')
output_audio_stream = output_container.add_stream("aac", rate=44100)

first_pts = None

next_pts = 0
for frame in input_container.decode(audio=0):

    # subtract first pts to start at zero
    if first_pts is None:
        first_pts = frame.pts
    frame.pts -= first_pts

    print(f"pts: {frame.pts}")
    print(f"time base: {frame.time_base}")
    print(f"sample rate: {frame.sample_rate}")
    print(f"samples: {frame.samples}")
    print(f"array_shape/2: {frame.to_ndarray().shape[1]/2}")  # we have stereo
    pts_per_sample = frame.time_base.denominator / frame.time_base.numerator
    pts_per_sample /= frame.sample_rate
    next_pts = frame.pts + pts_per_sample*frame.samples
    print(f"next pts: {next_pts}")
    print("---------")

    frame.pts = None
    output_container.mux(output_audio_stream.encode(frame))
```

Example output:
```
pts: 0
time base: 1/10000000
sample rate: 44100
samples: 4410
array_shape/2: 4410.0
next pts: 1000000.0
---------
pts: 980000
time base: 1/10000000
sample rate: 44100
samples: 4410
array_shape/2: 4410.0
next pts: 1980000.0
---------
pts: 2000000
time base: 1/10000000
sample rate: 44100
samples: 4410
array_shape/2: 4410.0
next pts: 3000000.0
---------
pts: 3010000
time base: 1/10000000
sample rate: 44100
samples: 4410
array_shape/2: 4410.0
next pts: 4010000.0
---------
pts: 4030000
time base: 1/10000000
sample rate: 44100
samples: 4410
array_shape/2: 4410.0
next pts: 5030000.0
---------
pts: 5050000
time base: 1/10000000
sample rate: 44100
samples: 4410
array_shape/2: 4410.0
next pts: 6050000.0
---------
```

How should we handle this?

fvollmer:
The pyav resampler and the fifo can't handle any pts inconsistencies, but ffmpeg can. If we take a look at the ffmpeg tool we can see that the heavy lifting for the resampling is done by the aformat filter (https://github.com/FFmpeg/FFmpeg/blob/169259d9a381a3c2132672da5c5f250fa194fb4d/fftools/ffmpeg_filter.c#L607). The fifo is implemented by the the buffersink (https://github.com/FFmpeg/FFmpeg/blob/169259d9a381a3c2132672da5c5f250fa194fb4d/fftools/ffmpeg_filter.c#L1113).

This PR to matches the pyav resampler behavior to the ffmpeg tool by using the same filters. This also simplifies the code quite a bit. 

Fixes #761
Depends on #765

### DIFF
--- a/av/audio/codeccontext.pyx
+++ b/av/audio/codeccontext.pyx
@@ -28,30 +28,22 @@ cdef class AudioCodecContext(CodecContext):
 
         cdef AudioFrame frame = input_frame
 
-        # Resample. A None frame will flush the resampler, and then the fifo (if used).
+        cdef bint allow_var_frame_size = self.ptr.codec.capabilities & lib.AV_CODEC_CAP_VARIABLE_FRAME_SIZE
+
         # Note that the resampler will simply return an input frame if there is
         # no resampling to be done. The control flow was just a little easier this way.
         if not self.resampler:
             self.resampler = AudioResampler(
-                self.format,
-                self.layout,
-                self.ptr.sample_rate
+                format=self.format,
+                layout=self.layout,
+                rate=self.ptr.sample_rate,
+                frame_size=None if allow_var_frame_size else self.ptr.frame_size
             )
-        frame = self.resampler.resample(frame)
+        frames = self.resampler.resample(frame)
 
-        cdef bint is_flushing = input_frame is None
-        cdef bint use_fifo = not (self.ptr.codec.capabilities & lib.AV_CODEC_CAP_VARIABLE_FRAME_SIZE)
-
-        if use_fifo:
-            if not self.fifo:
-                self.fifo = AudioFifo()
-            if frame is not None:
-                self.fifo.write(frame)
-            frames = self.fifo.read_many(self.ptr.frame_size, partial=is_flushing)
-            if is_flushing:
-                frames.append(None)
-        else:
-            frames = [frame]
+        # flush if input frame is None
+        if input_frame is None:
+            frames.append(None)
 
         return frames
 
--- a/av/audio/resampler.pxd
+++ b/av/audio/resampler.pxd
@@ -4,6 +4,7 @@ cimport libav as lib
 from av.audio.format cimport AudioFormat
 from av.audio.frame cimport AudioFrame
 from av.audio.layout cimport AudioLayout
+from av.filter.graph cimport Graph
 
 
 cdef class AudioResampler(object):
@@ -14,18 +15,12 @@ cdef class AudioResampler(object):
 
     cdef AudioFrame template
 
-    # Source descriptors; not for public consumption.
-    cdef unsigned int template_rate
-
     # Destination descriptors
     cdef readonly AudioFormat format
     cdef readonly AudioLayout layout
     cdef readonly int rate
+    cdef readonly unsigned int frame_size
 
-    # Retiming.
-    cdef readonly uint64_t samples_in
-    cdef readonly double pts_per_sample_in
-    cdef readonly uint64_t samples_out
-    cdef readonly bint simple_pts_out
+    cdef Graph graph
 
     cpdef resample(self, AudioFrame)
--- a/av/audio/resampler.pyx
+++ b/av/audio/resampler.pyx
@@ -1,13 +1,12 @@
 from libc.stdint cimport int64_t, uint8_t
 cimport libav as lib
 
-from av.audio.fifo cimport AudioFifo
-from av.audio.format cimport get_audio_format
-from av.audio.frame cimport alloc_audio_frame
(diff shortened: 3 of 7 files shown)


fvollmer:
Some comments on the modified tests:
`tests/test_audioresampler.py`:
 * the resampler returns now a list of frames, so we need to insert [0] if we want the first frame
 * the resampler can handle missing pts. If you want to treat them specially use async as you would have with the ffmpeg util
 * the resampler doesn't count the samples anymore. This now is handled by the filter. So testing samples_out is removed
 * if the time base is missing the new resampler will add it
 * I'm not sure what exactly a complex time base is, but the new resampler handles that test case fine. So I removed the exception test

`tests/test_codec_context.py`
 * adjust to handle a list of frames
 * remove `ctx.encode(None)` because it is already flushed (old resampler wouldn't always flush)

`tests/test_encode.py`
 * Depending on encoder `astream.frame_size` can be set. If it isn't we would get an audio frame with 0 samples, which wouldn't be useful

jlaine:
This looks very promising, and you are right: delegating the heavy lifting to FFmpeg code is definitely a win. My main concern here is that it's a breaking API change to `AudioResampler` (though I'm not sure how many people use it directly).

@mikeboers any houghts?

fvollmer:
Thanks for the feedback. If we really want to stay compatible we could put the new code into a `AudioResamplerv2` class, mark the `AudioResampler` class as deprecated and use the new v2 class in `codeccontext.pyx`.

But I'm not sure how many people are using the AudioResampler directly.

uvjustin:
Just a comment as a user of the library: I don't think breaking API changes are that bad as long as it's noted in release notes - PyAV is not updated very often and most users seem to have some development background and also probably don't have the library integrated into a system/updated through an automated process. ffmpeg itself moves quite fast, with major releases every 6-9 months, and there seem to be some API/ABI changes there relatively often, so even to just keep pace with ffmpeg I'd expect having some API changes here in PyAV.
I do sympathize with the sentiment that you maintainers @mikeboers and @jlaine have expressed - it's a beast to try to maintain everything across the whole library, and if the intention of the project was/is to satisfy your uses elsewhere, and it does that already, there's not much incentive to try and work out bugs/improve functionality elsewhere. That, and ffmpeg itself is a moving target, so it's hard to keep everything up to date.
Maybe the solution is to fork the library, or at least start a new branch, when ffmpeg does a new major release. Then PRs could be fast tracked with less regard for API stability until the bulk of the initial milestones are completed, and maybe some of the newer willing contributors could get more involved. ffmpeg 5.0 was released recently, so maybe now is an opportune time?

jlaine:
Thanks @fvollmer for this great contribution!

PR #767, opened 2021-03-28
Allow data stream without codec context as template when adding stream to OutputContainer.

jonasdehlin:
I have a use case where I need to update the PTS and DTS values of a KLV data stream. The suggested change will enable the use of the data stream from the input container as template when adding the new data stream to the output container.


jlaine:
Hi, could you please add a unit test for this?

jonasdehlin:
This is outdated and I will submit a new change request.

PR #772, opened 2021-04-04
Fix SideDataContainer Mapping semantics

jstrot:
Fixes Mapping's interfaces that try to index by value:

    TypeError: Type indices must be str, int, or Type


jstrot:
Moved my repo to https://github.com/qualIP/PyAV
Contact me when you're ready to accept code.

PR #775, opened 2021-04-12
Export ``AudioStream`` in ``av.audio.__init__``

cdce8p:
Currently `AudioStream` needs to be imported from `av.audio.stream` since it's missing from `av.audio.__init__`.
I noticed that `av.video.__init__` exports `VideoStream`, so doing the same for `AudioStream` seems logical.

https://github.com/PyAV-Org/PyAV/blob/main/av/video/__init__.py

### DIFF
--- a/av/audio/__init__.py
+++ b/av/audio/__init__.py
@@ -1,1 +1,2 @@
 from .frame import AudioFrame
+from .stream import AudioStream


KapJI:
@jlaine can you please take a look at this one? This is required when we want to add type annotations to functions which use PyAV.

jlaine:
+1 for consistency, thank you!

Issue #789, opened 2021-05-23
Allow setting ticks_per_frame


gsakkis:
`CodecContext.ticks_per_frame` is a read-only property, hardcoded to 1 when adding a new stream:

```py
>>> output = av.open("out.mp4", 'w')
>>> stream = output.add_stream('h264', 24)
>>> stream.ticks_per_frame
1
>>> stream.ticks_per_frame = 2
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
...
AttributeError: attribute 'ticks_per_frame' of 'av.codec.context.CodecContext' objects is not writable
```

It should be settable, most notably [H.264/MPEG-2 set it to 2](https://ffmpeg.org/doxygen/3.2/structAVCodecContext.html#a5c62b9c1528a272923ea2a4b86dea31a).

jlaine:
We seem to have two contradictory requests here. PR #790 advocates changing the way this member is not, but not exposing it. 

@rawler would making `ticks_per_frame` writable solve your issue too?

rawler:
Writeable would solve, but it would require a user of PyAV to be intimately familiar with these settings or one might get surprising results. `libvpx-vp9`, for example, seem to use this value in the bandwidth-calculations (how much time has passed between frames), so _not_ explicitly setting ticks_per_frame might break bandwidth-control quite badly. (IIRC, this is how we discovered the need to set it)

jlaine:
I'd argue that setting `ticks_per_frame` automatically qualifies the user as an "advanced" user. I've never once done this myself :)

Another thing that worries me is that I can't find any code in the FFmpeg CLI that sets "ticks_per_frame" outside the codecs and `avformat_transfer_internal_stream_timing_info`. Try grepping for "ticks_per_frame = ".

PR #790, opened 2021-05-27
codec/context: Defer ticks_per_frame-calculation

rawler:
Current hard-coded 1 `ticks_per_frame` assumes that `framerate =
1/time_base'. This is most often fine, except when re-encoding a source
with frames in non-fps intervals. Explicitly setting higher `time base`
works out OK for some codecs (libx264), but not with libvpx, which uses
`ticks_per_frame` and `time_base` to measure time for rate-control. The
net effect is that encoding with high-precision frame-timing for VP8 and
VP9 does not work in PyAV.


rawler:
Not sure about the failing test. It fails on "doc" generation, and I'm not how my change could have caused this error.

jlaine:
This looks like an interesting PR, could you rebase it on top of `main` to see if that shakes out the CI issue?

@mikeboers any opinion on this PR?

jlaine:
@rawler I've rebased your PR on top of the latest code. Could you explain a bit more where this logic is coming from, is there perhaps some equivalent code in the ffmpeg command line code?

av/container/output.pyx
```diff
@@ -100,7 +100,6 @@ cdef class OutputContainer(Container):
             codec_context.height = 480
             codec_context.bit_rate = 1024000
             codec_context.bit_rate_tolerance = 128000
-            codec_context.ticks_per_frame = 1
```
jlaine, line 103: Removing this seems fine, ticks_per_frame seems to default to 1 anyway

rawler:
The code was devised when we were transcoding to VP9, and noticed that the rate-control did not seem to work at all according to expectations. It seems the encoder uses the `ticks_per_frame`-field to figure out how much time has passed, and when encoding 25fps for webm (for example), the ticks are typically in ms, so 40 for each frame. With setting this field, we got it working.

Looking at ffmpeg, I'm not sure how it's solved there. I can't see `ticks_per_frame` being set anywhere. I believe it _may_ work by using `avcodec_open2` for all new encoder-contexts, which does not set `ticks_per_frame` but seems to rather [adjust `time_base`](https://ffmpeg.org/doxygen/4.1/libavcodec_2utils_8c_source.html#l01001) to match the default `ticks_per_frame`.

In either case, we've been running with this proposed patch in production for a while now. A colleague have found a corner-case in very large time-bases leading to int-overflow. Asking him now for permission to share.

jlaine:
See my comment in #789 : would it be sufficient to make ticks_per_frame writable?

rawler:
Sorry for very slow feedback. Had some time to look at this again today, and I now fail to repro the original case that failed for me. (Basically, bitrates going off the scales when encoding to VP9 and WebM, and incoming frames had high-precision time-stamps carried over.)

Not sure why, either this only happened for specific content (I've been testing with noise), or maybe newer versions of libvpx has stopped relying on the frame-duration so much. (`ticks-per-frame` is transparently forwarded as frame-duration to VPX)

In any case I guess we can close this PR, in favor or the alternate one (allowing setting ticks-per-frame would also allow different durations per-frame)

jlaine:
Thanks for reporting back!

Issue #629, opened 2020-04-22
UnicodeDecodeError with binary tags in mov files


Dobatymo:
## Overview

A UnicodeDecodeError is raised when opening some files.

## Expected behavior

It should not fail.

## Actual behavior

Traceback:
```
Traceback (most recent call last):
  File "asd.py", line 153, in __init__
    container = av.open(path, "r")
  File "av\container\core.pyx", line 365, in av.container.core.open
  File "av\container\input.pyx", line 70, in av.container.input.InputContainer.__cinit__
  File "av\utils.pyx", line 31, in av.utils.avdict_to_dict
  File "av\utils.pyx", line 14, in av.utils._decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

## Investigation

I noticed the problem when trying to open various `.mov` files. I think the problem might be the binary data in `com.apple.quicktime.artwork` tags.

## Reproduction

Just call `av.open("asd.mov", "r")`.
I have some example files, but cannot post them publicly.

## Versions

- OS: Windows 7 x64
- PyAV runtime:
```
PyAV v7.0.2.dev0
git origin: git@github.com:mikeboers/PyAV
git commit: v6.2.0-132-gd9bebbd
library configuration: --disable-static --enable-shared --enable-gpl --enable-version3 --enable-sdl2 --enable-fontconfig --enable-gnutls --e
nable-iconv --enable-libass --enable-libdav1d --enable-libbluray --enable-libfreetype --enable-libmp3lame --enable-libopencore-amrnb --enabl
e-libopencore-amrwb --enable-libopenjpeg --enable-libopus --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libtheora --enable-
libtwolame --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libzimg --enable
-lzma --enable-zlib --enable-gmp --enable-libvidstab --enable-libvorbis --enable-libvo-amrwbenc --enable-libmysofa --enable-libspeex --enabl
e-libxvid --enable-libaom --enable-libmfx --enable-amf --enable-ffnvcodec --enable-cuvid --enable-d3d11va --enable-nvenc --enable-nvdec --en
able-dxva2 --enable-avisynth --enable-libopenmpt
library license: GPL version 3 or later
libavcodec     58. 54.100
libavdevice    58.  8.100
libavfilter     7. 57.100
[...]

PR #807, opened 2021-09-02
Fix metadata decoding error

OndrejIT:
Hello,
this PR solved https://github.com/PyAV-Org/PyAV/issues/629

Maybe another fix is better? Or this is OK? ..


jlaine:
No this is not OK, this method must return a string, not "a string or bytes", the days of Python 2 are long gone :)

Issue #702, opened 2020-08-28
Allow user to force audio codec to be used instead of relying the default one


Kazhuu:
## Overview

Regarding [this](https://github.com/PyAV-Org/PyAV/issues/699) issue of mine earlier. Would it be possible to expose field `AvFormatContext::audio_codec_id` for user to force specific audio codec?

## Existing FFmpeg API

Documentation of the field [here](https://ffmpeg.org/doxygen/3.3/structAVFormatContext.html#ab179ade0dbe905ab2867b7001bf618ec).

## Expected PyAV API

Maybe add this to `av.open()` to be optional argument.

Example:
```
av.open('alsa-device', format='alsa', audio_codec='pcm_s32le')
```

I could also take this one and make contribution to this awesome library if this is considered something needed in general. Any thoughts?

paulhobbel:
Any news on the implementation of this?

Kazhuu:
Sorry but I have been busy with so many other things and haven't have time/interest for this one at the moment. Maybe someone else can give this a go instead. I tried to start implementing this but then soon realized that there are so many available codecs that left me wondering should all them be implemented or just selected few.

paulhobbel:
I'm currently trying out your solution but am failing to buil PyAV on my raspberry pi. Do you have any tips on how I can build PyAV?

Kazhuu:
I don't remember the steps out of my head but I think I was needing this package for one of my projects. I cloned it to my project and installed the package from there to my python virtual environment using `pipenv install -e ext/PyAV`. Then I just modified the sources in place. Of course you need to install Cython and other required packages to build the package. From readme the requirements are installed with `pip install --upgrade -r tests/requirements.txt`. Hope those helps.

Kazhuu:
@paulhobbel I made a PR of this [here](https://github.com/PyAV-Org/PyAV/pull/816)

Kazhuu:
I've a holiday coming weeks I could take a look at this.

Kazhuu:
ping

Kazhuu:
Is this still valid?

PR #816, opened 2021-09-26
Let user specified input audio codec and not relay on the default one

Kazhuu:
This PR relates to this #702 issue I posted earlier. This let user to specify input audio codec to be used instead of relaying on FFMPEG default one. Default codec is PCM_S16LE and in one scenario I needed to specify PCM_S32LE instead.

Some questions still on my head related to this:
* I only added PCM codecs. Is this okay? I think in the future if someone needs more they can be added quite easily.
* Anything to improve here? I'd be happy to modify this PR if needed.

With this change `av.open` can for example be called with
```
import av
av.open('alsa_device_name', format='alsa', audio_codec='PCM_S32LE')
```


jlaine:
Could we have some unit tests for this please?

av/container/core.pyx
```diff
@@ -167,6 +206,11 @@ cdef class Container(object):
         self.ptr.flags |= lib.AVFMT_FLAG_GENPTS
         self.ptr.max_analyze_duration = 10000000
 
+        if audio_codec is None:
```
jlaine, line 209: Input validation needs to be done much earlier. If you encounter a KeyError here, you're going to leak memory as we have already started allocating structures. I'm not even sure it makes sense to handle the None case, as I believe `AV_CODEC_ID_NONE` is already the default?

Kazhuu:
@jlaine do you think it would make sense to add documentation or something to list the supported codecs? I added a simple test but was thinking could I somehow test the default codec somehow when the `audio_codec` is not given. I couldn't find tests that open any alsa devices.

WyattBlue:
I'm closing this because this PR needs to be redone. It has merge conflicts and uses an old code style, (we prefer double quotes now).

I don't know if jlaine would agree with me but I think `av.open("alsa_device_name", format="alsa:pcm_s32le")` would be a better interface, (the parsing would be implemented in PyAV). I would think along these lines if you were to redo this.

For me, this is a very low priority feature. It's probably a low priority for jlaine too. I don't know anyone else who needs this. I wouldn't it if it were to be implemented, but it would need to be done in another PR.

Kazhuu:
Hello @WyattBlue. Thanks for commenting on this one. I don't really need this feature anymore so I don't see myself working on this PR again, also because the process has been quite slow and always forgot what I did last time. So this can be closed and the issue I opened also be closed. Maybe someone else who also needs this can find the continue the work. Thanks for you time!

PR #842, opened 2021-12-30
Simplify setup.py

jlaine:


jlaine:
@mikeboers I'm thinking of totally overhauling the setup.py to strip it down to its bare essentials, we have a *lot* of code in setup.py and I think Cython has matured a lot since our setup.py was written.

One major step forward is that we can *enforce* the presence of Cython, by adding a `pyproject.toml` containing:

```
[build-system]
requires = ["setuptools", "wheel", "Cython>=0.29.22"]
```

Any thoughts on this?

jlaine:
@WyattBlue I think I'm going to stop here for this PR.

I would eventually like to trim the setup.py down further but AFAICT this is already enough to stop importing distutils?

PR #870, opened 2022-01-04
Fix setting Stream time_base (fixes #784)

### LINKED ISSUE #784: Can't set video stream time_base
Hi, i have try to set video stream time_base with code bellow:
```
output = av.open("out.mp4", 'w')
stream = output.add_stream('h264', 24)
stream.pix_fmt = 'yuv420p'
stream.height = im_height
stream.width = im_width
stream.time_base = Fraction(1, 24)
print(stream.time_base)
print(stream.codec_context.time_base)
```
The command line output:
```
None
1/24
```
It means that the ```stream.time_base``` is not be assigned but ```stream.codec_context.time_base``` get it.
I use the container ```output``` to save a video, the ```time_base``` of packets is ```1/12288``` or other number, not ```1/24```.
I want to save a video that the ```time_base``` of ```AVStream``` set with me, how can i do ?

uvjustin:
In contrast to `__getattr__`, `__setattr__` [doesn't check the existing attributes](https://docs.python.org/3/reference/datamodel.html#object.__getattr__) first. This causes the `Stream.time_base` setter to never get called. This PR moves the `Stream.time_base` setter into `Stream.__setattr__`, and this should fix #784.

### DIFF
--- a/av/stream.pxd
+++ b/av/stream.pxd
@@ -24,6 +24,8 @@ cdef class Stream(object):
     # Private API.
     cdef _init(self, Container, lib.AVStream*)
     cdef _finalize_for_output(self)
+    cdef _set_time_base(self, value)
+    cdef _set_id(self, value)
 
 
 cdef Stream wrap_stream(Container, lib.AVStream*)
--- a/av/stream.pyx
+++ b/av/stream.pyx
@@ -127,6 +127,8 @@ cdef class Stream(object):
 
     def __setattr__(self, name, value):
         setattr(self.codec_context, name, value)
+        if name == "time_base":
+            to_avrational(value, &self._stream.time_base)
 
     cdef _finalize_for_output(self):
 
@@ -228,9 +230,6 @@ cdef class Stream(object):
         """
         def __get__(self):
             return avrational_to_fraction(&self._stream.time_base)
-
-        def __set__(self, value):
-            to_avrational(value, &self._stream.time_base)
 
     property average_rate:
         """
--- a/av/stream.pyx
+++ b/av/stream.pyx
@@ -126,9 +126,12 @@ cdef class Stream(object):
                 raise AttributeError(name)
 
     def __setattr__(self, name, value):
+        if name == "id":
+            self._set_id(value)
+            return
         setattr(self.codec_context, name, value)
         if name == "time_base":
-            to_avrational(value, &self._stream.time_base)
+            self._set_time_base(value)
 
     cdef _finalize_for_output(self):
 
@@ -195,11 +198,14 @@ cdef class Stream(object):
         def __get__(self):
             return self._stream.id
 
-        def __set__(self, v):
-            if v is None:
-                self._stream.id = 0
-            else:
-                self._stream.id = v
+    def _set_id(self, value):
+        """
+        Setter used by __setattr__ for id property.
+        """
+        if value is None:
+            self._stream.id = 0
+        else:
+            self._stream.id = value
 
     property profile:
         """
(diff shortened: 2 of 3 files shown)


uvjustin:
Also noticed `Stream.id` has the same issue, so refactored the PR slightly and changed the `id` setter. Since `CodecContext` has a matching `time_base` but not an `id`, setting the former also sets the same attribute on the `CodecContext`, while setting the latter does not.

jlaine:
I'm not sure I'm comfortable with this approach, I really dislike our `__setattr__` / `__getattr__` which encourage confusing a stream and a codec context. In an ideal world I think I'd like them to go away entirely.

Could you summarise what the root cause of #784 is?

uvjustin:
So the problem is that stream already uses `__getattr__` and `__setattr__` to pass things through to the `CodecContext` (and to the Codec in the case of `__getattr__`).
However, as detailed in the link I included above, `__getattr__`/`__setattr__` have an asymmetry - when getting an attribute, `__getattr__` will actually only get called if an existing attribute of the given name does not exist. However when setting an attribute, `__setattr__` will swallow the request and the setter of the given name (if it exists) will NOT be called.
This is the root cause of #784 - the setters for `id` and `time_base` never get called.
I agree with you on the confusing nature of the class hierarchy. After I made this PR I actually did a refactor to create a separate `CodecContext` for each PyAV `Stream` as AVStream.codec is no longer exposed. However I ended it scrapping it as I realized there is quite a bit of drift between the PyAV codebase and ffmpeg, and a more well thought out refactor of the class hierarchy might be necessary.

jlaine:
OK, I think I get it.. so `__setattr__` is really masking *all* our setters.

I'm wondering whether our goal shouldn't be to drop `__setattr__` and `__getattr__` completely, there is really too much magic going on. What I don't know is how we can have an intermediate step where we issue a deprecation warning inviting users to update their code to directly reach into `stream.codec_context` themselves if that's what they mean.

Regarding `time_base` specifically I see your patch makes the setter change *both* the stream and the codec's time_base, is this the expected behaviour?

uvjustin:
Yes, exactly. I agree with you too...passing through different attributes makes it unclear what is getting get/set on where. I'm not sure the best way to address this.
Yes, that was what I intended to do but I'm not sure if that's what makes sense. I'm not too clear myself on what the class/attribute hierarchy should look like. My thinking was people would only want to set a time_base on an output container's stream, and that would mean setting it on the `CodecContext` as well. The setting on the `Stream` itself is just so that you can see that same value the next time there's a getter request for the attribute. Storing the state in two places like this doesn't seem very clean.

uvjustin:
Just to circle back on this, in ffmpeg `time_base` is stored in both `AVStream` and in `AVCodecContext`
For `AVCodecContext`, the comments read:

> encoding: MUST be set by user.
> decoding: the use of this field for decoding is deprecated. Use framerate instead.

, while for `AVStream`, the comments read:

> decoding: set by libavformat
> encoding: May be set by the caller before avformat_write_header() to provide a hint to the muxer about the desired timebase. In avformat_write_header(), the muxer will overwrite this field with the timebase that will actually be used for the timestamps written into the file (which may or may not be related to the user-provided one, depending on the format).

As you noted, we can deprecate the use of `__getattr__` and `__setattr__`, forcing getting and setting on the Codec/CodecContexts directly, but actually removing that functionality will probably be quite a breaking change (working on the Stream is easier than on the Codec/CodecContext and adds quite a lot of convenience, so I assume it is broadly used).

However, the current code is just wrong as it includes two setters that are never actually used. Maybe the best thing is to just remove them and issue a warning when they are used.

jlaine:
OK this is getting too deep, I'll merge the PR as-is.

PR #872, opened 2022-01-05
Replace deprecated av_init_packet with av_packet_alloc

uvjustin:
[`av_init_packet` is deprecated as of FFmpeg 4.4](https://github.com/FFmpeg/FFmpeg/blob/9a194252c78c74c4455f837fd152b61904f5dc85/doc/APIchanges#L224). This PR replaces `av_init_packet` with `av_packet_alloc` and uses `av_packet_free` to both free the `AVPacket` and clear the `AVPacket*` (The replacement functions [have been available since FFmpeg 3.0](https://github.com/FFmpeg/FFmpeg/blob/9a194252c78c74c4455f837fd152b61904f5dc85/doc/APIchanges#L1125)).

### DIFF
--- a/av/container/output.pyx
+++ b/av/container/output.pyx
@@ -210,18 +210,17 @@ cdef class OutputContainer(Container):
         self.start_encoding()
 
         # Assert the packet is in stream time.
-        if packet.struct.stream_index < 0 or <unsigned int>packet.struct.stream_index >= self.ptr.nb_streams:
+        if packet.ptr.stream_index < 0 or <unsigned int>packet.ptr.stream_index >= self.ptr.nb_streams:
             raise ValueError('Bad Packet stream_index.')
-        cdef lib.AVStream *stream = self.ptr.streams[packet.struct.stream_index]
+        cdef lib.AVStream *stream = self.ptr.streams[packet.ptr.stream_index]
         packet._rebase_time(stream.time_base)
 
         # Make another reference to the packet, as av_interleaved_write_frame
         # takes ownership of it.
-        cdef lib.AVPacket packet_ref
-        lib.av_init_packet(&packet_ref)
-        self.err_check(lib.av_packet_ref(&packet_ref, &packet.struct))
+        cdef lib.AVPacket *packet_ptr = lib.av_packet_alloc()
+        self.err_check(lib.av_packet_ref(packet_ptr, packet.ptr))
 
         cdef int ret
         with nogil:
-            ret = lib.av_interleaved_write_frame(self.ptr, &packet_ref)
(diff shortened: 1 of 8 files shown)


jlaine:
This is awesome, thanks so much for putting in the work!

From a cosmetic point of view, I don't see any occurrences of a "struct_ptr" in our current code, it seems to be "ptr" in all the cases I can see. Any thoughts on this?

av/container/output.pyx
```diff
@@ -210,18 +210,17 @@ cdef class OutputContainer(Container):
...
-        cdef lib.AVStream *stream = self.ptr.streams[packet.struct.stream_index]
+        cdef lib.AVStream *stream = self.ptr.streams[packet.struct_ptr.stream_index]
         packet._rebase_time(stream.time_base)
 
         # Make another reference to the packet, as av_interleaved_write_frame
         # takes ownership of it.
-        cdef lib.AVPacket packet_ref
-        lib.av_init_packet(&packet_ref)
-        self.err_check(lib.av_packet_ref(&packet_ref, &packet.struct))
+        cdef lib.AVPacket *packet_ptr = lib.av_packet_alloc()
```
jlaine, line 220: Couldn't we write this as:

`packet_ref = lib.av_packet_clone(packet.struct_ptr)`?

uvjustin:
You're right, let's just change it to ptr. I'll push a commit

uvjustin:
@jlaine Weird, I got a comment from you about using `lib.av_packet_clone()` but I can't see it here now.
Anyway, yes, I think we could just use `lib.av_packet_clone()`. But then do we need to use a different kind of err_check? `lib.av_packet_clone()` returns a NULL pointer on error. I'm not sure if there's a specific error checking framework we are supposed to use in this repo. Are allocations supposed to be checked? For example, I didn't check the return from `lib_av_packet_alloc()` here, but I see that `lib.av_frame_alloc()` in `frame.pyx` is not checked either.

jlaine:
> @jlaine Weird, I got a comment from you about using `lib.av_packet_clone()` but I can't see it here now. Anyway, yes, I think we could just use `lib.av_packet_clone()`. But then do we need to use a different kind of err_check? `lib.av_packet_clone()` returns a NULL pointer on error. I'm not sure if there's a specific error checking framework we are supposed to use in this repo. Are allocations supposed to be checked? For example, I didn't check the return from `lib_av_packet_alloc()` here, but I see that `lib.av_frame_alloc()` in `frame.pyx` is not checked either.

Yeah I pulled my comment precisely because I realised that `av_packet_clone` doesn't check the return code, so I don't really know what to think. Anyway, heads up: I force pushed on your branch to squash the commits, I'm happy to merge as-is once CI completes.

PR #877, opened 2022-01-23
Avoid unnecessary vsnprintf() calls in log_callback()

mephi42:
`mov_read_trak()` and `mov_read_trun()` generate a lot of `AV_LOG_TRACE`
logs, which are ultimately thrown away. However, `log_callback()` still
formats and processes them. This causes `av_log()` to - at least in some
cases I observed with perf record - to take 97% of
`avformat_open_input()` run time. Improve this by returning from
`log_callback()` early when a log is obviously not needed.

### DIFF
--- a/av/logging.pyx
+++ b/av/logging.pyx
@@ -245,6 +245,12 @@ cdef void log_callback(void *ptr, int level, const char *format, lib.va_list arg
     if not inited and not print_after_shutdown:
         return
 
+    # Fast path: avoid logging overhead. This should match the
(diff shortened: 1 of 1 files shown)


jlaine:
This sounds like a good idea in general but I have some questions:

- if the filtering is already performed in `log_callback`, why do it as well in `log_callback_gil`?
- your filtering is not identical to the existing one, why the `level != lib.AV_LOG_ERROR` special case?

jlaine:
Thanks, merged!

PR #881, opened 2022-01-24
[tests] test against FFmpeg 4.4

jlaine:


tests/test_encode.py
```diff
@@ -60,7 +60,11 @@ def assert_rgb_rotate(self, input_):
     self.assertIsInstance(stream, VideoStream)
     self.assertEqual(stream.type, 'video')
     self.assertEqual(stream.name, 'mpeg4')
-    self.assertEqual(stream.average_rate, 24)  # Only because we constructed is precisely.
+    # Depending on the FFmpeg version the `average_rate` seems to differ:
+    # - FFmpeg 4.4 reports Fraction(1152, 47)
+    # - Ffmpeg < 4.4 reports exactly 24
+    self.assertGreaterEqual(stream.average_rate, 24)
+    self.assertLess(stream.average_rate, 25)
     self.assertEqual(stream.rate, Fraction(24, 1))
```
hmaarrfk, line 68:
```suggestion
    self.assertGreaterEqual(stream.rate, 24)
    self.assertLess(stream.rate, 25)
```

hmaarrfk:
I'm sorry to jump in, but you don't think this kind of error is an "off by one" error?:

```
1152 / 48 = 24
and 
1152 % 47 = 24
```

The latest error to show up also seems suspiciously close to an off by one error.

jlaine:
@hmaarrfk something about the FFmpeg internals seems to have changed starting from 4.4, I really don't know what.

jlaine:
I've purposefully removed the special handling as something doesn't feel right. I'm starting to think something needs to be fixed in our code, I really don't see why the framerate should be different. The only thing that would convince me that the code is right and the test are wrong is if we replicate this behaviour using the FFmpeg command line tool.

hmaarrfk:
Ok great. That is my thinking too. I just wanted to make sure it was correct. 

Do you think it has to do with things similar to:
https://github.com/PyAV-Org/PyAV/pull/905/files

Where the fraction rounding isn't entirely correct due to floating point rounding errors?

jlaine:
I don't think rounding comes into play at all here, we are creating a video stream at 24 fps with 48 frames.

Also I have just tried generating a video using ffmpeg (4.4.1, Debian) on the command line from a set of PNG images (00.png -> 47.png):

`ffmpeg -r 24 -i %02d.png foo.mp4`

The resulting video has a frame rate of *exactly* 24fps, so I do think we have an actual issue.

Issue #886, opened 2022-02-14
How do I send EOF to a av.filter.Graph?


FirefoxMetzger:
## Overview

I'm trying to use the FFMPEG filter `palettegen` to compute a color pallette for a video (and then turn it into a GIF later on). However, the filter expects to read a full video and wishes to receive an EOF error ([relevant line](https://github.com/FFmpeg/FFmpeg/blob/d15f65c4699de54eafc3b9a8f145cad978666520/libavfilter/vf_palettegen.c#L564) inside FFmpeg source) before it outputs a single 256 color pallette.

When using the binary, this can be achieved as: `ffmpeg -i testsrc.mp4 -vf palettegen pallette.png`
(To generate a test video: `ffmpeg -f lavfi -i testsrc -t 10 -pix_fmt yuv444p testsrc.mp4`)

However, writing the same in pyAV the filter never produces an output, because it waits for the EOF signal. I don't know how to send it, nor have I been able to work it out from the source code. A pointer would be highly appreciated :)

Here is the code as far as I got:

```python
import av
import av.filter

graph = av.filter.Graph()

with av.open("testsrc.mp4", "r") as container:
    video_stream = container.streams.video[0]
    
    # setup filter graph
    video_in = graph.add_buffer(template=video_stream)
    palette_gen_filter = graph.add("palettegen")
    video_out = graph.add("buffersink")
    video_in.link_to(palette_gen_filter)
    palette_gen_filter.link_to(video_out)
    graph.configure()

    for frame in container.decode(video=0):
        graph.push(frame)

    # TODO: Send EOF ... but how?

    palette_frame = graph.pull()  # this currently fails   

```

The above currently fails with the following exception:

```
Traceback (most recent call last):
  File "delete_me.py", line 22, in <module>
    palette_frame = graph.pull()  # this currently fails
  File "av/filter/graph.pyx", line 212, in av.filter.graph.Graph.pull
  File "av/filter/context.pyx", line 107, in av.filter.context.FilterContext.pull
  File "av/error.pyx", line 336, in av.error.err_check
av.error.BlockingIOError: [Errno 11] Resource temporarily unavailable
```

FirefoxMetzger:
I found the solution: It is currently impossible to do so.

[...]

PR #887, opened 2022-02-24
Enable sending EOF to Graphs

FirefoxMetzger:
Closes: #886 

This PR adds the ability to send EOF by pushing None on a video `buffer` as proposed in #886 

@jlaine I don't know if you have a preferred code style for the repo, so I didn't apply anything. Let me know if this needs more additions from your side or what is missing to make this complete :)

### DIFF
--- a/av/filter/graph.pyx
+++ b/av/filter/graph.pyx
@@ -196,12 +196,14 @@ cdef class Graph(object):
 
     def push(self, frame):
 
-        if isinstance(frame, VideoFrame):
+        if frame is None:
+            contexts = self._context_by_type.get('buffer', []) + self._context_by_type.get('abuffer', [])
+        elif isinstance(frame, VideoFrame):
             contexts = self._context_by_type.get('buffer', [])
         elif isinstance(frame, AudioFrame):
             contexts = self._context_by_type.get('abuffer', [])
         else:
-            raise ValueError('can only push VideoFrame or AudioFrame', type(frame))
+            raise ValueError('can only AudioFrame, VideoFrame or None; got %s' % type(frame))
 
         if len(contexts) != 1:
             raise ValueError('can only auto-push with single buffer; found %s' % len(contexts))
--- a/av/filter/context.pyx
+++ b/av/filter/context.pyx
@@ -77,7 +77,10 @@ cdef class FilterContext(object):
 
     def push(self, Frame frame):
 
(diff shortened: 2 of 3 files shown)


av/filter/graph.pyx
```diff
@@ -196,7 +196,11 @@ cdef class Graph(object):
 
     def push(self, frame):
 
-        if isinstance(frame, VideoFrame):
+        if frame is None:
+            contexts = self._context_by_type.get('buffer', [])
+            if len(contexts) == 0:
```
jlaine, line 201: I'm not to sure about this, if we receive None, we really have no way of knowing whether it's for audio or for video.. Would it make sense to flush both?

FirefoxMetzger: @jlaine I agree; we can't know. I chose this based on the assumption that `graph.push` and `graph.pull`  are only meant for use with only one source and one sink total. Perhaps I misunderstood? 

If there are multiple sources of the same kind, `push` will raise a ValueError

https://github.com/PyAV-Org/PyAV/blob/95b4f818c0b53c18af8a1ae0d5d9044df1586545/av/filter/graph.pyx#L206-L207

and similarly pull will raise if there is more than one sink in total:

https://github.com/PyAV-Org/PyAV/blob/95b4f818c0b53c18af8a1ae0d5d9044df1586545/av/filter/graph.pyx#L216-L218

That said, I'm happy to push None to both contexts instead if there are multiple, or just leave pushing of None out of graphs (you can still use the context directly) if you think that is cleaner. Let me know.

jlaine: OK you're right, this "push" method only handles a single buffer. Maybe use the following:

`contexts = self._context_by_type.get('buffer', [])  + self._context_by_type.get('abuffer', [])`

That way it will bomb in the same way as for AudioFrame / VideoFrame?

jlaine:
I've force-pushed to your branch with the suggested change along with a linter fix

jlaine:
@fvollmer I have a feeling this PR has some implications on your resampler changes (which I have merged), shouldn't we make use of it to flush out the resampler?

FirefoxMetzger:
> I've force-pushed to your branch with the suggested change along with a linter fix

Thanks for the heads up. I'm also happy with the changes you made, so from my side there isn't anything more to add.

jlaine:
And it's merged, thanks a lot!

PR #910, opened 2022-03-15
Add support for FFmpeg 5.0

jlaine:


jlaine:
@mikeboers This is progressing better than I thought. One interesting aspect is that we now hit the exact same failures as the ones encountered against FFmpeg 4.4.

uvjustin:
@jlaine Using a separate `AVCodecContext` for PyAV is the same approach I took when I attempted a refactor. I thought there would be more issues with PyAV using a `AVCodecContext` distinct from the internal one held by `AVStream`, but it doesn't seem like this causes too many problems.
It does seem like framerate in `AVCodecContext` is not passed through `AVCodecParameters` when copying back and forth, and there seem to be a few different framerate/time_base fields between `AVCodecContext` and `AVStream`, and I think we need to take care to maintain consistency there.
BTW, the subtitles test issue with ffmpeg 5.0 seems to be related to the deprecation of the old ASS format with timing here: https://ffmpeg.org/pipermail/ffmpeg-cvslog/2016-February/098534.html .

av/container/output.pyx
```diff
@@ -101,11 +97,7 @@ cdef class OutputContainer(Container):
...
             codec_context.ticks_per_frame = 1
-
-            rate = Fraction(rate or 24)
-
-            codec_context.framerate.num = rate.numerator
-            codec_context.framerate.den = rate.denominator
+            to_avrational(rate or 24, &codec_context.framerate)
 
             stream.avg_frame_rate = codec_context.framerate
             stream.time_base = codec_context.time_base
```
uvjustin, line 103: I realize this was in the existing code, but is codec_context.time_base ever set before here?

jlaine: I *think* time_base gets initialized to a sane value, but it would be worth double checking. I really appreciate you looking at this PR, feel free to tinker and suggest changes as I'm not 100% sure of what I'm doing! When in doubt I look at how FFmpeg (CLI) does it.

jlaine:
NOTE: I noticed there is an issue in `close_output`, because Container.streams can already be partially destroyed, so calling it from OutputContainer.__dealloc__ is not reliable. I'm not sure I can do much than emitting a warning if this happens.

I've probably also introduced a memory leak, I need to free the CodecContext.

av/container/output.pyx
uvjustin: Sure, I will also tinker with it a bit when I get a chance. It doesn't seem too far off. I think some of the failing tests may actually be the result of existing mismatches between PyAV and ffmpeg. It's hard to get everything to align fully as some of the fields like framerate can get set by ffmpeg in a number of different places depending on the container/codec used.

uvjustin:
For the subtitles test failing on 5.0, I think we can just change the test to also accept the regular ASS format without the "Dialogue: " prefix and the two timestamps (basically just allow what it's failing on now).

jlaine:
> For the subtitles test failing on 5.0, I think we can just change the test to also accept the regular ASS format without the "Dialogue: " prefix and the two timestamps (basically just allow what it's failing on now).

No I don't think that's the issue. I've tried changing the format, which does strip away the timings from the output. However I think it masks the underlying issue which is that the timings look plain wrong.

jlaine:
> The current ASS format doesn't include the timings in the output though. Are you checking them in `SubtitleProxy`? Edit: If you are referring to the 0.97 and 2.54 subtitle times embedded in the ASS string, those are only in the deprecated ASS format, but we can check them by adding:
> 
> ```
>         self.assertIsInstance(subs[0], SubtitleSet)
>         self.assertEqual(subs[0].pts, 970000)
>         self.assertEqual(subs[0].start_display_time, 0)
>         self.assertEqual(subs[0].end_display_time, 1570)
> ```

OK you're exactly right, so I have reworked the tests as suggested to explicitly check timings. I've also made "ass" format the default instead of "ass_with_timings" to get consistent behaviour across FFmpeg versions.

That leaves the other failures related to rates.

uvjustin:
@jlaine This looks pretty good and the tests seem to be passing.
I'll make a build against ffmpeg 5.0 and test it against our usage in Home Assistant.

jlaine:
I had put off merging this to see if any bug-fixes needed to land in the 9.x branch, as it will be the last one supporting FFmpeg < 4.3. It looks as though we're in a good place, so we're close to the point where this PR should land.

One thing that bothers me is that we're hardly testing anything at all on Windows (i.e. only FFmpeg 4.3). Conda seems to have dropped the ball on packaging FFmpeg for Windows..

uvjustin:
Maybe we can try to use the Windows binaries from https://github.com/BtbN/FFmpeg-Builds/releases instead of from Conda?

av/container/output.pyx
```diff
@@ -64,14 +65,12 @@ cdef class OutputContainer(Container):
...
-            codec = codec_obj.ptr
 
         else:
-            if not template._codec:
-                raise ValueError("template has no codec")
-            if not template._codec_context:
+            if not template.codec_context:
                 raise ValueError("template has no codec context")
-            codec = template._codec
+            codec_obj = template.codec_context
```
uvjustin, line 72:
```suggestion
            codec_obj = template.codec_context.codec
```

av/container/input.pyx
```diff
@@ -65,7 +70,18 @@ cdef class InputContainer(Container):
 
         self.streams = StreamContainer()
         for i in range(self.ptr.nb_streams):
-            self.streams.add_stream(wrap_stream(self, self.ptr.streams[i]))
+            stream = self.ptr.streams[i]
+            codec = lib.avcodec_find_decoder(stream.codecpar.codec_id)
+            if codec:
+                # allocate and initialise decoder
+                codec_context = lib.avcodec_alloc_context3(codec)
+                err_check(lib.avcodec_parameters_to_context(codec_context, stream.codecpar))
```
uvjustin, line 78: `avcodec_parameters_to_context` does not set the `framerate` field of `AVCodecContext`: https://www.ffmpeg.org/doxygen/trunk/codec__par_8c_source.html#l00182
As `Stream.rate`/`Stream.framerate` relies on getattr pass through to its codec_context's rate/framerate, this means `Stream.rate`/`Stream.framerate` are no longer valid for input video streams.
Two ways we can remedy this:
1) We can set the codec_context's framerate here and continue to rely on this passthrough feature.
2) We can add `rate`/`framerate` properties directly to `VideoStream`.
Note that `AudioStream` does not have this issue as the `sample_rate` field gets copied properly.

.github/workflows/tests.yml
```diff
@@ -149,9 +147,6 @@ jobs:
       matrix:
         config:
           - {os: windows-latest, python: 3.7, ffmpeg: "4.3"}
```
uvjustin, line 149: I've added FFmpeg 4.4 to the matrix in #997, and that PR now uses our own pyav-ffmpeg library for testing. But if we want to test on Windows against FFmpeg 5.0, we can use the same method with a third-party build at least until our own 5.x builds are complete.

uvjustin (changes requested):
Looks pretty good. I left some comments with a few things I noticed when getting a stopgap release ready on the ha-av fork.

av/container/output.pyx
jlaine: Ehm. So this basically means we don't have any test coverage for this..

.github/workflows/tests.yml
jlaine: Awesome I've merged your PR and rebased this one

av/container/input.pyx
jlaine: I'm not even sure we should be doing anything:

https://www.ffmpeg.org/doxygen/trunk/structAVCodecContext.html#a4d08b297e97eefd66c714df4fff493c8

`decoding: For codecs that store a framerate value in the compressed bitstream, the decoder may export it here. { 0, 1} when unknown.`

uvjustin: I'm pretty sure it's not always set. It was causing a failure in our test suite that I remedied using the 2nd option above.
However, what you said actually implies we should rely on the first option above (passthrough), because the decoder may export it when decoding the bitstream.

Issue #738, opened 2021-02-03
How do I use os.pipe() to feed data to an av container?


zhengwei-coding:
## Overview
Dear all, i am trying to use pyav to decode a local file with a customized format. I uses pipe's write-end to feed the data in while pipe's read-end file object to pass to av.open(file) for decoding. However, it seems av.open(read_pipe) always throw TypeError.

My code and actual behaver is as follow:
>>> import os, av      
>>> r, w = os.pipe()
>>> fr = os.fdopen(r, 'rb')
>>> fw = os.fdopen(w, 'wb')
>>> ct = av.open(fr) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "av\container\core.pyx", line 354, in av.container.core.open
  File "av\container\core.pyx", line 117, in av.container.core.Container.__cinit__
TypeError: File's name attribute must be string-like.
>>>

I am wondering if it just doesn't support a file object with the name being a number (fdopen seems to return a file object which's name is the fd (type int instead of string) or i did anything wrong here...   

Thanks for the help!!

zhengwei-coding:
Hi all, followed the traceback info, i found the code line 113 ~ 117 in av/container/core.pyx as below:
        if isinstance(file_, str):
            self.name = file_
        else:
            self.name = getattr(file_, 'name', '<none>')
            if not isinstance(self.name, str):
                raise TypeError("File's name attribute must be string-like.")
            self.file = file_

it seems that it will check if the name attribute is a type of string (normally it is). However, in case the file is opened with its file descriptor, the name will be assigned as its file descriptor which is an integer. 
I am wondering if this type check is necessary here or should contain the case with # at least. 
Any comment?

Cheers.

jlaine:
I agree this check is overly restrictive, file-like objects don't necessarily have a name, for instance `io.BytesIO` has no name at all..

farleylai:
Any follow-up plan to fix the issue?
The stdout from `subprocess.Popen(...)` also does not have the name of `str` but `int` implying it's from some fileno that should be general to accept.

jlaine:
Sure, write a PR including unit tests.

farleylai:
Got it. Will definitely consider that.
So far in the short term, the `name` property from a pipe may be modified by accessing the underlying `raw.__dict__` as follows:

```python
    proc = subprocess.Popen(['./cmd-to-run'], stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
[...]

PR #920, opened 2022-03-24
Make it possible to read from a pipe (fixes: #738)

jlaine:


jlaine:
@mikeboers I've been thinking this one over quite a bit, and I'm wondering: do you have any file-like objects in mind which *do not* inherit [IOBase](https://docs.python.org/3/library/io.html#io.IOBase) ?

The reason I'm asking is that:
- the presence of `read()` does not guarantee the object is readable, we also need `readable()` to return `True`
- the presence of `write()` does not guarantee the object is writable, we also need `writable()` to return `True`
- the presence of `seek()` and `tell()` does not guarantee the object is seekable, we also need `seekable()` to return `True`

=> the question becomes "do we mandate the presence of `XXXable()`? If not, how do we interpret the absence of such a method? The "optimistic" (assume it's OK) approach retains backwards compatibility.

PR #923, opened 2022-03-25
[buffer] deprecate Buffer.to_bytes()

jlaine:
This method is no longer needed since with Python 3 we can simply use
bytes(buf).


jlaine:
@mikeboers are you OK with this?

mikeboers (approved):
I am on board.

PR #947, opened 2022-03-28
Add CIBW_REPAIR_WHEEL_COMMAND_MACOS to workflow

uvjustin:
Add CIBW_REPAIR_WHEEL_COMMAND_MACOS to fix wheel building issues when compiling with libvpx support.
See https://github.com/PyAV-Org/pyav-ffmpeg/pull/51 for context.


jlaine:
I don't think I agree with this, I don't see why vpx should require special treatment?

uvjustin:
Let me see if we can do this another way

Issue #909, opened 2022-03-13
Python interpreter hangs on exit when using 'AUTO' thread_type on Linux


Breakthrough:
## Overview

Setting `thread_type = 'AUTO'` prevents the Python interpreter from gracefully exiting and can hang indefinitely.  I have reproduced the issue with PyAV version 8 and 9, on Ubuntu 16.04, 18.04, and 20.04 (see reproduction below).  I also believe the issue affects [the threading example in the cookbook](https://pyav.org/docs/stable/cookbook/basics.html#threading), which makes no mention of the possible issues nor workarounds that exist.

I have not been able to reproduce the issue on Windows, and I am unable to test this on OSX.

## Expected behavior

PyAV should not prevent the program from terminating when using `AUTO` threading mode, or if this is unavoidable, the threading example in the cookbook should better indicate that this can happen and how it can be mitigated.

The [caveats section](https://pyav.org/docs/stable/overview/caveats.html#sub-interpeters) implies this issue only affects subinterpreters, but I'm not sure that's the case.  I'm also curious if it would be possible to disable logging by default when PyAV is shutting down which would likely resolve most user facing issues.

## Actual behavior

Any application using PyAV on Linux can occasionally hang on termination when using `thread_type = 'AUTO'` so long as at least one frame has been decoded.

## Reproduction

Verified using the following minimum working example:

```python
import av
print('Starting test.')
container = av.open('video.mp4')
container.streams.video[0].thread_type = 'AUTO'
# Have to decode at least one frame to trigger the bug.
next(container.decode(video=0))
print('Test complete.')

# No longer hangs after uncommenting the following line:
#av.logging.restore_default_callback()
```

Running this in a loop via bash will output something like:

```
$ while true; do python3 test_av_hang.py; done
Starting test.
Test complete.
Starting test.
Test complete.
Starting test.
Test complete.
```

The output will then stop suddenly.

I discovered the issue on Travis CI, and was able to reproduce locally in both a VM and on an actual Ubuntu distro.  I had difficulty reproducing the issue with only a single-core VM, but it seems to fail rather quickly on any multicore Linux machine.  Adding `av.logging.restore_default_callback()` at the end of the program resolves the issue.

This issue also occurs with [the threading example in the cookbook](https://pyav.org/docs/stable/cookbook/basics.html#threading), which makes no mention of this problem.

[...]

PR #948, opened 2022-03-28
Ensure `close` is called when Container is deleted

Breakthrough:
Currently, close() is only called when using a context manager. This can demonstrably lead to hangs when using AUTO threading mode as shown in the cookbook example.

As @jlaine discovered in #909, calling `close()` on the Container seems to resolve the issue, likely by making sure the threads have been joined by the time __dealloc__ runs.

### DIFF
--- a/av/container/core.pyx
+++ b/av/container/core.pyx
@@ -244,6 +244,9 @@ cdef class Container(object):
             # Finish releasing the whole structure.
             lib.avformat_free_context(self.ptr)
 
+    def __del__(self):
+        self.close()
+        
     def __enter__(self):
         return self
 


Breakthrough:
Note that `__exit__` can probably be made a no-op after this change, but I wasn't sure if anything else relied on that behaviour so left it as-is for now.  Let me know if you want that fixed up as well.

jlaine:
I just tried this out and it makes no difference, in fact I don't even see __del__ being closed?

Breakthrough:
Sorry I didn't realize, something must be keeping a reference to the container internally within the library, so `__del__` never gets called even when the object is explicitly deleted with `del`.  For my own use case it's sufficient to just call `close()` manually in a `__del__` method, but might be worth investigating what could be causing this for PyAV.  Not a huge priority though - thanks again for all your help with this.

jlaine:
You're welcome! Now that you've gained some insights into PyAV's internals, do drop by now and again. I could do with some additional hands triaging bugs or investigating .. :)

PR #959, opened 2022-04-12
Free packet used for muxing in OutputContainer

uvjustin:
https://github.com/PyAV-Org/PyAV/pull/872 seems to have introduced a memory leak as the packet that is allocated in `mux_one` is not freed. This PR adds a packet_ptr to the OutputContainer which follows the OutputContainer's lifecycle and is freed in its `__dealloc__`.

### DIFF
--- a/av/container/output.pyx
+++ b/av/container/output.pyx
@@ -34,9 +34,13 @@ cdef class OutputContainer(Container):
     def __cinit__(self, *args, **kwargs):
         self.streams = StreamContainer()
         self.metadata = {}
+        with nogil:
+            self.packet_ptr = lib.av_packet_alloc()
 
     def __dealloc__(self):
         close_output(self)
+        with nogil:
+            lib.av_packet_free(&self.packet_ptr)
 
     def add_stream(self, codec_name=None, object rate=None, Stream template=None, options=None, **kwargs):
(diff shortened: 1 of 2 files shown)


av/container/output.pyx
```diff
@@ -220,10 +224,9 @@ cdef class OutputContainer(Container):
...
         # takes ownership of it.
-        cdef lib.AVPacket *packet_ptr = lib.av_packet_alloc()
-        self.err_check(lib.av_packet_ref(packet_ptr, packet.ptr))
+        self.err_check(lib.av_packet_ref(self.packet_ptr, packet.ptr))
 
         cdef int ret
         with nogil:
-            ret = lib.av_interleaved_write_frame(self.ptr, packet_ptr)
+            ret = lib.av_interleaved_write_frame(self.ptr, self.packet_ptr)
         self.err_check(ret)
```
uvjustin, line 232: lib.av_interleaved_write_frame will reinitialize self.packet_ptr, so we can reuse it on subsequent calls: https://ffmpeg.org/doxygen/trunk/group__lavf__encoding.html#ga37352ed2c63493c38219d935e71db6c1

jlaine:
Ouch. This looks good to me, if you could squash your commits and add your explanation to the commit message I'd be happy to merge it.

Issue #951, opened 2022-03-31
Encoding unit tests fails if libx264 not found


deepsworld:
## Overview

The `python setup.py test` script fails if FFMPEG is built with openh264 instead of libx264 codec. Since the conda-forge distribution of FFMPEG is GPL enabled, I am trying to avoid that infection for pyav by building it with openh264.  It will be great if pyav supports both and either for these kind of situations. 

## Actual behavior

```log
ERROR: test_encoding_with_pts (tests.test_encode.TestBasicVideoEncoding)
Traceback (most recent call last):
  File "/home/ml/dpatel/miniconda3/envs/build39/conda-bld/av_1648738792808/work/tests/test_encode.py", line 113, in test_encoding_with_pts
    stream = output.add_stream("libx264", 24)
  File "av/container/output.pyx", line 62, in av.container.output.OutputContainer.add_stream
    codec_obj = codec_name if isinstance(codec_name, Codec) else Codec(codec_name, 'w')
  File "av/codec/codec.pyx", line 184, in av.codec.codec.Codec.__cinit__
    self._init(name)
  File "av/codec/codec.pyx", line 193, in av.codec.codec.Codec._init
    raise UnknownCodecError(name)
av.codec.codec.UnknownCodecError: libx264
```

Build report:
```log
======================================================================
ERROR: test_encoding_with_pts (tests.test_encode.TestBasicVideoEncoding)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ml/dpatel/miniconda3/envs/build39/conda-bld/av_1648738792808/work/tests/test_encode.py", line 113, in test_encoding_with_pts
    stream = output.add_stream("libx264", 24)
  File "av/container/output.pyx", line 62, in av.container.output.OutputContainer.add_stream
    codec_obj = codec_name if isinstance(codec_name, Codec) else Codec(codec_name, 'w')
  File "av/codec/codec.pyx", line 184, in av.codec.codec.Codec.__cinit__
    self._init(name)
  File "av/codec/codec.pyx", line 193, in av.codec.codec.Codec._init
    raise UnknownCodecError(name)
av.codec.codec.UnknownCodecError: libx264

----------------------------------------------------------------------
Ran 208 tests in 9.807s

FAILED (errors=1, skipped=2)
Test failed: <unittest.runner.TextTestResult run=208 errors=1 failures=0>
error: Test failed: <unittest.runner.TextTestResult run=208 errors=1 failures=0>
[...]

PR #962, opened 2022-04-18
[tests] use "h264" instead of "libx264" for tests (fixes: #951)

jlaine:
This allows running the test suite against FFmpeg builds which do not
have libx264 enabled.


jlaine:
@deepsworld this one is for you. Changing "libx264" to "h264" doesn't seem to break the test suite *if libx264 is available*. I also tried removing libx264 support from our FFmpeg, but then things do start to break

jlaine:
I think I'm most likely going to close this issue, it doesn't actually solve anything..

hmaarrfk:
I think you need a few more patches:
https://github.com/conda-forge/av-feedstock/blob/main/recipe/lgpl_tests.patch

hmaarrfk:
For more clarity, we at a conda-forge now build ffmpeg with (gpl) and without (lgpl) x264. 

We then compile pyav with lgpl, and then test it with the gpl and the lgpl versions. Without the patch I linked to, the lgpl-only tests would fail.

hmaarrfk:
Nevermind, it seems that your patch is good. It actually includes the addition 5 line cleanup that I didn't include in mine due to not wanting to touch too many lines of code.

Did you build your FFMPEG builds with openh264? I'm happy to test the patch at conda-forge (where we have lgpl builds readily available).

hmaarrfk (approved):

PR #963, opened 2022-04-18
[package] update FFmpeg binaries

jlaine:
Fix vpx flags for macos, see https://github.com/PyAV-Org/pyav-ffmpeg/pull/63


jlaine:
@uvjustin this makes use of the FFmpeg builds with your fix. Am I good to merge ?

uvjustin:
Good to merge, thanks

Issue #805, opened 2021-09-01
Clarification about PyAV license?


bjuncek:
## Overview

License file of `PyAV` is advertised as [BSD3](https://github.com/PyAV-Org/PyAV/blob/main/LICENSE.txt); however, as in conda-forge it is packaged directly with (non-free) version of FFMPEG, shouldn't the binaries should be advertised as GPL license as well? 

## Expected behavior

Whilst the codebase is and can be BSD3 licensed, it should be noted for the user that installation of the binaries makes the entire package GPL-licensed. 
For example, `conda-forge` FFMPEG is [clearly marked](https://github.com/conda-forge/ffmpeg-feedstock) as being GPL as a binary in their feedstock. OpenCV on another hand has a [note](https://github.com/opencv/opencv-python/blob/c1cc7e5652d2a55e0bf603a9e021f5300a99caee/LICENSE-3RD-PARTY.txt#L243) in their license file that user takes responsibility for non-free FFMPEG as it's redistributed with it's packages.

## Actual behavior

Licensing is at the moment mentioned in the license file linked above, in the [docs](https://pyav.org/docs/stable/development/license.html?highlight=license), and conda-forge/pypi feedstocks respectively. There is no mentioning of a usage of non-free FFMPEG, or GPL license. 

When submitting bug report, I found the mention in the collection script (see output bellow), but I think it should be clearly marked in the repo/feedstocks as well. 

```
PyAV v8.0.3
git origin: git@github.com:PyAV-Org/PyAV
git commit: unknown-commit
[ ... ]
library license: GPL version 2 or later
libavcodec     58. 54.100
[ ... ]
```

## Investigation

FFMPEG the is mostly, *LGPL* (thus allowing dynamic linking without license change), but FFMPEG built in conda-forge (and the one that is installed by default when one installs pyav from the `conda-forge`), is built with `--enable-gpl` flag set, making it *GPL* license (see [here](https://www.ffmpeg.org/legal.html) for details) that _does not allow dynamic linking_. 

As PyAV links to the GPL version of FFMPEG, unless I misunderstand the way PyAV links to FFMPEG, that the binaries should be advertised as such, either in the repository directly (see opencv example above as an option) or in the pypi/conda-forge directly. 

I have done the following:

- [x] Checked the [PyAV documentation](https://pyav.org/docs)
- [x] Searched on [Google](https://www.google.com/search?q=pyav+how+do+I+foo)
[...]

PR #967, opened 2022-04-21
Add FFmpeg licensing information

bjuncek:
Per #805, I've added licensing information to `LICENSE.txt` and (perhaps unnecessary) to `README.md`.
The information added to `README.md` would be useful to have in PyPI and conda-forge pages, but I'm unsure on how to add this.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -26,5 +26,5 @@ EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ------------------------------------------------------------------------------
 
 FFmpeg that is redistributed within all PyAV wheels and binaries is licensed
-under GNU General Public License (GPL) version 2 or later. If those binaries
-are used, this license applies to all of PyAV. 
+under their respective license outlined below:
+http://ffmpeg.org/legal.html
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -21,3 +21,10 @@ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
 OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
 NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
 EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+------------------------------------------------------------------------------
+
+FFmpeg that is redistributed within all PyAV wheels and binaries is licensed
+under GNU General Public License (GPL) version 2 or later. If those binaries
+are used, this license applies to all of PyAV. 
--- a/README.md
+++ b/README.md
@@ -56,6 +56,14 @@ pip install --upgrade -r tests/requirements.txt
 make
 ```
 
+Licensing
+----------
+
+PyAV python package is available under [BSD3 license](LICENSE.txt).
+[PyPI][pypi] and [conda-forge][conda-forge] binaries are shipped with FFmpeg
+binaries under GPLv2 or later.
+
+
 ---
 
 Have fun, [read the docs][docs], [come chat with us][gitter], and good luck!
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ Licensing
 ----------
 
 PyAV python package is available under [BSD3 license](LICENSE.txt).
-[PyPI][pypi] and [conda-forge][conda-forge] binaries are shipped with FFmpeg
-binaries under GPLv2 or later.
+[PyPI][pypi] binaries are shipped with FFmpeg
+binaries under their respective [license](http://ffmpeg.org/legal.html). 
 
 
 ---


README.md
```diff
@@ -56,6 +56,14 @@ pip install --upgrade -r tests/requirements.txt
 make
 ```
 
+Licensing
+----------
+
+PyAV python package is available under [BSD3 license](LICENSE.txt).
+[PyPI][pypi] and [conda-forge][conda-forge] binaries are shipped with FFmpeg
```
jlaine, line 63: We are not in charge of the conda-forge binaries, I really have no idea what they link FFmprg against (especially as they seem to use third party binaries for Windows), so I'd rather keep that out of our license files.

LICENSE.txt
```diff
@@ -21,3 +21,10 @@ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
 OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
 NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
 EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+------------------------------------------------------------------------------
+
+FFmpeg that is redistributed within all PyAV wheels and binaries is licensed
+under GNU General Public License (GPL) version 2 or later. If those binaries
```
jlaine, line 29: How do we know that the GPL V2 is the one that applies as FFmpeg is linked against many libraries with different licenses?

jlaine:
Thanks so much for looking into this! I'm not sure I follow how we can determine "the" license for the binary wheels distribution, as it is a collection of several softwares with different licenses?

README.md
bjuncek: Makes sense - didn't know that, sorry.

jlaine: No worries, we do mention the conda installation method in our docs, we should maybe make it clearer this is a distribution outside our control.

LICENSE.txt
jlaine: I think we really need to look at what other projects distributing wheels with FFmprg are doing. I'm feeling somewhat out of my depth on this topic.

rgommers:
> Thanks so much for looking into this! I'm not sure I follow how we can determine "the" license for the binary wheels distribution, as it is a collection of several softwares with different licenses?

The way to deal with this is to list all the licenses of bundled components. For example, what SciPy does is:
1. Have separate file content if it varies for different wheels: https://github.com/MacPython/scipy-wheels/blob/master/LICENSE_win32.txt
2. Amend the license file as part of the wheel build process: https://github.com/MacPython/scipy-wheels/blob/master/patch_code.sh#L12-L17

bjuncek:
@jlaine Would you be ok with the solution that Ralf proposed? 
Specifically, creating a separate LICENSE file for FFMPEG (any other binary info can be included in there), that would be appended to the license file in the makefile when wheels are built?

In terms of determining THE licence, GPL [FAQ](https://www.gnu.org/licenses/gpl-faq.html#IfInterpreterIsGPL) states that code that utilises and makes calls to a GPL-licensed software becomes GPL:

> Linking [name of your program] statically or dynamically with other modules is making a combined work based on [name of your program]. Thus, the terms and conditions of the GNU General Public License cover the whole combination

hmaarrfk:
Unsolicited comment from a maintainer of the conda-forge ffmpeg and pyav feedstocks:

Do remember that the license of the binary work can be different than the license of the constituent source files. I also don't know of many court cases that have been tried for the GPL license so it is somewhat unclear as to how it applies in practice (also i am not a lawyer so none of this is legal advice)

FFMPEG is somewhat interesting, it assumes that you don't have a license for x264, and that you are using the GPL'ed version. You can purchase one from https://x264.org/en/ so it somewhat gets tricky. 

Ultimately, many projects have simply copied over the license files of the source projects, and left it to the end user to untangle them in terms of what it means for the final project. (Un)fortunately, the LGPL process explicitly allows for this. They want you to be able to change the underlying library which you linked to, to one that is potentially under a different license.

For example, at conda-forge, I've made it quite explicit that during the build process [1] of PyAV, we build with the LGPL version, the PyAV compiled files are not really bound by the GPL license anymore. The user is able to then choose which version of ffmpeg they want to install, the GPL compiled one (which includes x264 and x265 among other things) or the non-gpl compiled one. What this does to their resulting license of PyAV is unclear to me. 

My ask, is that you continue to differentiate between the license of the source (which, according to the history of the license file, you did not intend on being licensed through the GPL license), and the license of the binary, which may, according to different interpretations, be subject to the GPL terms.

In terms of the pypi distribution, I also feel like it would be nice for your end users, if you didn't set the license to GPL (but that is up to you). I'm not sure about the best way to do this on pypi. Optional packages come to mind, but I can't comment on the details of the implementation. A different license for the `sdist` and the `wheel` might be an other option.

[1] https://github.com/conda-forge/av-feedstock/pull/60

WyattBlue:
The appended readme message in this PR is vague and not very helpful. Our binaries are GPLv3 in the PyPI release (with other compatible licenses for deps), but this git repo is BSD by itself, and can be used to build other licenses (LGPLv2, LGPLv3, GPLv2).

Our `LICENSE.txt` should NOT be modified, since this repo does not contain any code GPL or LGPL code.

I'm going to close this, since this PR has major issues.

PR #981, opened 2022-05-09
Fix useful_array when unaligned and dtype > 1 byte

corco:
This fix the function useful_array when unaligned and dtype's size is 2 or more bytes. See the added test that fails with previous' function.

The issue is that the current fonction crops the numpy array using sizes in bytes instead of in number of array elements. When dtype is not a byte, the wrong cropping occurs.

There are several way to fix the function, such as expressing `total_line_size` and `useful_line_size` in number of elements instead of bytes. I choose to crop the numpy array in bytes before reinterpreting it in `dtype`.

### DIFF
--- a/av/video/frame.pyx
+++ b/av/video/frame.pyx
@@ -66,10 +66,10 @@ cdef useful_array(VideoPlane plane, unsigned int bytes_per_pixel=1, str dtype='u
     import numpy as np
     cdef size_t total_line_size = abs(plane.line_size)
     cdef size_t useful_line_size = plane.width * bytes_per_pixel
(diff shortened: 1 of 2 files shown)


jlaine:
This looks like a good catch, thanks! Let's let CI run but I'm happy with the changes, especially as they come with a unit test

corco:
Can you merge this please? I have another PR that depends on this :P

jlaine:
Merged, thanks! I am swamped with work ATM so I have great trouble keeping up with PyAV + aiortc + aioquic + etc..

PR #997, opened 2022-06-19
Add CI test for Win ffmpeg 4.4

uvjustin:
Use ffmpeg binary from ~https://github.com/BtbN/FFmpeg-Builds/releases~ pyav-ffmpeg.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -148,6 +148,7 @@ jobs:
       fail-fast: false
       matrix:
         config:
+          - {os: windows-latest, python: 3.7, ffmpeg: "4.4"}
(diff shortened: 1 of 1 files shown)


jlaine:
Hm. using third-party binary builds has bitten us in the backside in the past. I'm wondering if we should just use our own builds from pyav-ffmpeg if we cannot rely on conda anymore?

uvjustin:
Yeah, we could just do that too. It does seem that something is wrong with the Windows conda builds.
FWIW the repo used is linked off the `[ffmpeg.org webpage](http://ffmpeg.org/download.html#build-windows)`,

uvjustin:
Updated to use the pyav-ffmpeg binary

jlaine:
Looks good to me, can you squash your commits please?

Issue #1005, opened 2022-07-06
VideoStream.rate and VideoStream.framerate no longer available


uvjustin:
## Overview

After #910 was merged, VideoStream.rate and VideoStream.framerate are no longer populated from an InputContainer.

## Expected behavior

These fields should be populated.

## Investigation

Using the below test_streams.py file (augment the current test_streams.py with 2 additional lines), the test case now fails. Pre #910 it succeeded.

```
import av

from .common import TestCase, fate_suite


class TestStreams(TestCase):
    def test_stream_tuples(self):

        for fate_name in ("h264/interlaced_crop.mp4",):

            container = av.open(fate_suite(fate_name))

            video_streams = tuple([s for s in container.streams if s.type == "video"])
            self.assertEqual(video_streams, container.streams.video)

            audio_streams = tuple([s for s in container.streams if s.type == "audio"])
            self.assertEqual(audio_streams, container.streams.audio)

    def test_selection(self):

        container = av.open(fate_suite("h264/interlaced_crop.mp4"))
        video = container.streams.video[0]
        # audio_stream = container.streams.audio[0]
        # audio_streams = list(container.streams.audio[0:2])

        self.assertEqual([video], container.streams.get(video=0))
        self.assertEqual([video], container.streams.get(video=(0,)))
        self.assertIsNotNone(video.rate)
        self.assertIsNotNone(video.framerate)

        # TODO: Find something in the fate suite with video, audio, and subtitles.
```

## Actual behavior

fails

## Research

I have done the following:

- [ ] Checked the [PyAV documentation](https://pyav.org/docs)
- [ ] Searched on [Google](https://www.google.com/search?q=pyav+how+do+I+foo)
- [ ] Searched on [Stack Overflow](https://stackoverflow.com/search?q=pyav)
- [ ] Looked through [old GitHub issues](https://github.com/PyAV-Org/PyAV/issues?&q=is%3Aissue)
- [ ] Asked on [PyAV Gitter](https://gitter.im/PyAV-Org)
- [ ] ... and waited 72 hours for a response.

## Additional context

This is because the InputContainer's cinit was changed to initialize a separate AVCodecContext, and the parameters are copied over from the stream's AVCodecParameters using lib.avcodec_parameters_to_context(). However, AVCodecParameters does not include a video framerate, so the AVCodecContext never has the framerate field initialized for video streams. Note that AudioStream does not have this issue as the sample_rate field gets copied properly.
[...]

PR #1007, opened 2022-07-07
Add rate and framerate properties to VideoStream (fixes: #1005)

uvjustin:
Fixes the regression introduced by #910 

This PR fixes the issue by adding properties that point directly to the underlying AVStream's avg_frame_rate. It doesn't set the rate on the new codec_context, but that doesn't appear to be necessary.

### DIFF
--- a/tests/test_encode.py
+++ b/tests/test_encode.py
@@ -121,7 +121,7 @@ def assert_rgb_rotate(self, input_, is_dash=False):
     self.assertEqual(stream.format.name, "yuv420p")
     self.assertEqual(stream.format.width, WIDTH)
     self.assertEqual(stream.format.height, HEIGHT)
-    self.assertEqual(stream.codec_context.rate, None)
+    self.assertEqual(stream.rate, expected_average_rate)
     self.assertEqual(stream.ticks_per_frame, 1)
 
 
--- a/tests/test_encode.py
+++ b/tests/test_encode.py
@@ -121,7 +121,7 @@ def assert_rgb_rotate(self, input_, is_dash=False):
     self.assertEqual(stream.format.name, "yuv420p")
     self.assertEqual(stream.format.width, WIDTH)
     self.assertEqual(stream.format.height, HEIGHT)
-    self.assertEqual(stream.rate, None)
+    self.assertEqual(stream.codec_context.rate, None)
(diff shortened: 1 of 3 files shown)


tests/test_encode.py
```diff
@@ -121,7 +121,7 @@ def assert_rgb_rotate(self, input_, is_dash=False):
     self.assertEqual(stream.format.name, "yuv420p")
     self.assertEqual(stream.format.width, WIDTH)
     self.assertEqual(stream.format.height, HEIGHT)
-    self.assertEqual(stream.rate, None)
+    self.assertEqual(stream.codec_context.rate, None)
```
jlaine, line 124: No, please update all tests to explicitly test both stream.rate and stream.codec_context.rate

uvjustin: The original line seems to have been added in #910 and I don't know if it is even necessary?

jlaine: It is definitely necessary since apparently our test suite did not capture the behaviour you are proposing to fix!

uvjustin: What I mean is the previous assertion that it is None doesn't seem to make sense. Same with the updated assertion.
We should be asserting that stream.rate is Fraction(24,1), and I can make that change. I can add assertions on the codec_context too (there seems to be just one other place), but note that we don't do the same for the audio streams currently.

uvjustin: > It is definitely necessary since apparently our test suite did not capture the behaviour you are proposing to fix!

Just asserting stream.rate is Fraction(24,1) alone does capture the behavior...

uvjustin: Yikes, added that test and hit the problem from #881

jlaine:
I'm having another look at this, and I'm really confused. We already have `Stream.average_rate`, which does the exact same thing but with a more explicit name. Why do we need two additional accessors?

uvjustin:
I see your point and that makes sense. We should just probably remove these accessors, but we should be explicit that it is a breaking change.

uvjustin:
I guess a follow on question is what to do with AudioStream.
The actual rate field for an audio stream is stored in sample_rate of the AVCodecParameters. We should probably get rid of the `rate` passthrough and enforce `sample_rate` since it is a more explicit name as well.

jlaine:
Closing in favour of #1036

PR #1030, opened 2022-09-22
Add CodecContext.opt_set_int() and CodecContext.opt_get_int()

sirf:
Expose methods to manipulate options in CodecContext.

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -647,3 +647,16 @@ cdef class CodecContext(object):
 
         def __set__(self, value):
             self.ptr.skip_frame = SkipType[value].value
+
+    # Default search_flags to AV_OPT_SEARCH_CHILDREN
+    def opt_set_int(self, name: str, value: int, search_flags: int = 1):
+        ret = lib.av_opt_set_int(self.ptr, name, value, search_flags)
+        if ret != 0:
+            raise RuntimeError(f'av_opt_set_int() = {ret}')
+
+    def opt_get_int(self, name: str, search_flags: int = 1):
+        cdef int64_t val
+        ret = lib.av_opt_get_int(self.ptr, name, search_flags, &val)
+        if ret != 0:
+            raise RuntimeError(f'av_opt_get_int() = {ret}')
+        return val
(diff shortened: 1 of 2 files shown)


sirf:
I want to manipulate options while CodecContext is open, i.e. after `avcodec_open2` has been called. Is that possible with CodecContext.options or through some other existing method? If so then awesome!
Any help would greatly appreciated.

jlaine:
Ah, updating options after the codec has been opened is another matter, we definitely don't support it right now as our options are only passed when we call `avcodec_open2()`.

Do you know if there is a "high level" API to update options? My concern is that I'd rather not add all possible "opt_set_XXX" calls if there is a higher level API.

jlaine:
I think `av_opt_set_dict()` might be what we are looking for.

We could for instance have an `update_options()` method which would allow us to update multiple options at once, regardless of their type. I'm not too sure of the exact API, as we already have a "self.options", and it's not clear to me what we should do with it.

sirf:
It would be neat if CodecContext.options was a custom dict-like object that calls `av_opt_set` when `__setitem__` is called, and calls `av_opt_set_dict` when `update()` is called :)

jlaine:
> It would be neat if CodecContext.options was a custom dict-like object that calls `av_opt_set` when `__setitem__` is called, and calls `av_opt_set_dict` when `update()` is called :)
> 
> Would of course also work on all the other AV* structs that has options.

Sure that would be great, I'm just not too sure:

- whether `av_opt_set_dict` "patches" existing options or fully replaces them
- how we can have this "magic" options dict in Cython

Still, I think it would be worth you experimenting with some form of API (e.g. an `update_options()` method) to check whether you are actually able to change `crf` as you desire.

sirf:
How about this then:

A method named `update_options` that has the exact same semantics as `dict.update` / `collections.abc.MutableMapping.update`, e.g. you can update one or many options from an existing dict, update one or many options through keyword arguments, from a list of tuples, etc.
```
opts = { "crf": "20", "some_other_option": "123" }
codec.update_options(opts)

opts = { "crf": "20", "some_other_option": "123" }
codec.update_options(opts, third_option="456")

codec.update_options(crf="20")

codec.update_options(some_option="1", some_other_option="123")

codec.update_options([ ("some_option", "1"), ("some_other_option", "123") ])
```

I did an implementation and updated the code in this PR. It works, and it avoids the creation of an intermediary `AVDictionary` that `av_opt_set_dict` would have needed.

PR #1034, opened 2022-10-16
Remove extraneous rate attributes

uvjustin:
This PR removes the following attributes:
AudioCodecContext.rate (use AudioCodecContext.sample_rate instead)
AudioFrame.rate (use AudioFrame.sample_rate instead)
VideoStream.rate and VideoStream.framerate (use VideoStream.average_rate instead)
VideoCodecContext.rate (use VideoCodecContext.framerate instead)

### DIFF
--- a/av/audio/codeccontext.pyx
+++ b/av/audio/codeccontext.pyx
@@ -75,14 +75,6 @@ cdef class AudioCodecContext(CodecContext):
         def __set__(self, int value):
             self.ptr.sample_rate = value
 
-    property rate:
-        """Another name for :attr:`sample_rate`."""
-        def __get__(self):
-            return self.sample_rate
-
-        def __set__(self, value):
-            self.sample_rate = value
-
     # TODO: Integrate into AudioLayout.
     property channels:
         def __get__(self):
--- a/av/audio/frame.pyx
+++ b/av/audio/frame.pyx
@@ -165,14 +165,6 @@ cdef class AudioFrame(Frame):
         def __set__(self, value):
             self.ptr.sample_rate = value
 
-    property rate:
-        """Another name for :attr:`sample_rate`."""
-        def __get__(self):
-            return self.ptr.sample_rate
-
-        def __set__(self, value):
-            self.ptr.sample_rate = value
-
     def to_ndarray(self, **kwargs):
         """Get a numpy array of this frame.
 
--- a/tests/test_encode.py
+++ b/tests/test_encode.py
@@ -121,7 +121,6 @@ def assert_rgb_rotate(self, input_, is_dash=False):
     self.assertEqual(stream.format.name, "yuv420p")
     self.assertEqual(stream.format.width, WIDTH)
     self.assertEqual(stream.format.height, HEIGHT)
-    self.assertEqual(stream.rate, None)
     self.assertEqual(stream.ticks_per_frame, 1)
 
 
@@ -129,13 +128,13 @@ class TestBasicVideoEncoding(TestCase):
     def test_default_options(self):
         with av.open(self.sandboxed("output.mov"), "w") as output:
             stream = output.add_stream("mpeg4")
+            self.assertEqual(stream.average_rate, Fraction(24, 1))
             self.assertEqual(stream.bit_rate, 1024000)
             self.assertEqual(stream.format.height, 480)
             self.assertEqual(stream.format.name, "yuv420p")
             self.assertEqual(stream.format.width, 640)
             self.assertEqual(stream.height, 480)
             self.assertEqual(stream.pix_fmt, "yuv420p")
-            self.assertEqual(stream.rate, Fraction(24, 1))
             self.assertEqual(stream.ticks_per_frame, 1)
             self.assertEqual(stream.time_base, None)
             self.assertEqual(stream.width, 640)
@@ -185,7 +184,7 @@ class TestBasicAudioEncoding(TestCase):
             stream = output.add_stream("mp2")
             self.assertEqual(stream.bit_rate, 128000)
             self.assertEqual(stream.format.name, "s16")
-            self.assertEqual(stream.rate, 48000)
+            self.assertEqual(stream.sample_rate, 48000)
             self.assertEqual(stream.ticks_per_frame, 1)
             self.assertEqual(stream.time_base, None)
 
@@ -262,7 +261,7 @@ class TestEncodeStreamSemantics(TestCase):
                     # decoder didn't indicate constant frame size
                     frame_size = 1000
                 aframe = AudioFrame("s16", "stereo", samples=frame_size)
-                aframe.rate = 48000
+                aframe.sample_rate = 48000
                 apackets = astream.encode(aframe)
                 if apackets:
                     apacket = apackets[0]
--- a/av/audio/resampler.pyx
+++ b/av/audio/resampler.pyx
@@ -94,7 +94,7 @@ cdef class AudioResampler(object):
             if (
                 frame.format.sample_fmt != self.template.format.sample_fmt or
                 frame.layout.layout != self.template.layout.layout or
-                frame.sample_rate != self.template.rate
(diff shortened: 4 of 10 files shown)


av/audio/codeccontext.pyx
```diff
@@ -75,14 +75,6 @@ cdef class AudioCodecContext(CodecContext):
         def __set__(self, int value):
             self.ptr.sample_rate = value
 
-    property rate:
```
jlaine, line 78: I'm fine with this, let's stick to the exact FFmpeg terminology.

av/audio/frame.pyx
```diff
@@ -165,14 +165,6 @@ cdef class AudioFrame(Frame):
         def __set__(self, value):
             self.ptr.sample_rate = value
 
-    property rate:
```
jlaine, line 168: Same as above, agreed.

av/video/stream.pyx
```diff
@@ -10,3 +10,13 @@ cdef class VideoStream(Stream):
             self.codec_context.height,
             id(self),
         )
+
+    def __getattr__(self, name):
```
jlaine, line 14: Uhoh, getattr / setattr have already bitten us many times, why do we need this?

tests/test_encode.py
```diff
@@ -135,7 +134,7 @@ def test_default_options(self):
             self.assertEqual(stream.format.width, 640)
             self.assertEqual(stream.height, 480)
             self.assertEqual(stream.pix_fmt, "yuv420p")
-            self.assertEqual(stream.rate, Fraction(24, 1))
+            self.assertEqual(stream.average_rate, Fraction(24, 1))
```
jlaine, line 137: Can we move this so that attributes are sorted alphabetically?

jlaine:
On the whole I agree with this PR as it moves use closer to FFmpeg's terminology and removes some confusion about the different "rates". Could you maybe improve the commit message with the motivation for these changes?

av/video/stream.pyx
uvjustin: This blocks pass through access to `framerate` from the existing getattr/setattr methods in the Stream class. We can also do this by making an explicit `framerate` property here instead of overriding the getattr/setattr methods.
However, maybe we can just remove the passthrough access altogether. It's a larger breaking change, but I just tried it and it doesn't seem to break the tests. What do you think?

uvjustin: Whoops I lied, it doesn't work

av/video/stream.pyx
```diff
@@ -10,3 +10,12 @@ cdef class VideoStream(Stream):
             self.codec_context.height,
             id(self),
         )
+
+    property framerate:
+        """Explicitly block pass through access to the framerate property."""
+        def __get__(self):
+            raise AttributeError("framerate")
+
+    cdef _set_framerate(self, value):
```
jlaine, line 19: I really don't understand this, I've never seen a setter declared like this? Is there some Cython magic at play here?

av/video/stream.pyx
jlaine: I'm really not comfortable with adding any more `__setattr__` or `__getattr__` dunder methods. If the root cause of our issue is the existing dunder methods in `Stream`, yes by all means block access there with an explicit comment stating why we are doing it. A unit test which checks an AttributeError is raised on get/set seems necessary too.

av/video/stream.pyx
uvjustin: I don't know actually - that's just how it was done with the existing properties.

av/video/stream.pyx
uvjustin: Sure, that sounds fine to me. I'll just remove any special treatment. However, that means #1005 still stands, and any existing code that relies on VideoStream.framerate may silently fail.

av/video/stream.pyx
jlaine: No,  the general pattern is that you have a __set__ at the same level as your __get__, or the more modern syntax is just with @property like regular python ;)

uvjustin: LOL yeah, you're right. It was late where I am and I actually copied the syntax from the _set_id workaround code for setattr (which was actually from my own work before :-P). Anyway it's irrelevant now since this section has been removed.

av/video/stream.pyx
uvjustin: How about raising an `AttributeError` when accessing `framerate` in the existing getattr/setattr of `Stream`? That avoids adding more dunder methods and avoids the issues with adding a separate property.

jlaine:
I have put up an alternative proposal in #1035 which only deprecates the aliases instead of removing them.

av/video/stream.pyx
uvjustin: I'll open that as a separate PR so we can evaluate it separately.

uvjustin:
> I have put up an alternative proposal in #1035 which only deprecates the aliases instead of removing them.

I agree deprecation is generally friendlier, but I think the problem of #1005 comes back again. VideoStream.rate may no longer be populated properly as it was before. In this case I'd rather the library throw an explicit error than merely printing a deprecation warning.

jlaine:
> > I have put up an alternative proposal in #1035 which only deprecates the aliases instead of removing them.
> 
> I agree deprecation is generally friendlier, but I think the problem of #1005 comes back again. VideoStream.rate may no longer be populated properly as it was before. In this case I'd rather the library throw an explicit error than merely printing a deprecation warning.

I'm getting kind of uncomfortable here, for a couple of reasons:

- We're talking about pushing through breaking changes shortly before a release. I don't have any idea how this will impact our users.
- Pulling or deprecating the "rate" properties is not directly related to #1005. If you were using `.framerate` directly, you're not getting any extra help here, this property's behaviour has changed.
- After reading some more code, the "rate" (with no reference to frames or samples) terminology crops up in a couple more places: `add_stream` takes a rate argument, `average_rate` doesn't say explicitly it's a *frame* rate (FFmpeg's avg_frame_rate does..) 
- We are proposing blocking access to `Stream.framerate`, but `Stream.sample_rate` seems very much required. Where does this difference come from?

uvjustin:
> 
> I'm getting kind of uncomfortable here, for a couple of reasons:
> 
> * We're talking about pushing through breaking changes shortly before a release. I don't have any idea how this will impact our users.

#910 already changed the behavior which is a breaking change. I know because it broke my usage of the library. This just explicitly breaks things so that users don't have to dig in to find out where the problem is coming from.

> * Pulling or deprecating the "rate" properties is not directly related to [VideoStream.rate and VideoStream.framerate no longer available #1005](https://github.com/PyAV-Org/PyAV/issues/1005). If you were using `.framerate` directly, you're not getting any extra help here, this property's behaviour has changed.

Yes, that's why I originally removed framerate here and now have opened #1036.

> * After reading some more code, the "rate" (with no reference to frames or samples) terminology crops up in a couple more places: `add_stream` takes a rate argument, `average_rate` doesn't say explicitly it's a _frame_ rate (FFmpeg's avg_frame_rate does..)

> * We are proposing blocking access to `Stream.framerate`, but `Stream.sample_rate` seems very much required. Where does this difference come from?

See the "Additional Context" section of #1005. ffmpeg treats sample_rate differently from framerate. https://www.ffmpeg.org/doxygen/trunk/structAVCodecParameters.html

jlaine:
OK thanks for the clarification regarding sample_rate. It seems to me we'll at some point need to make up our minds whether we should be providing direct access to `codecpar` or to the codec context, but that sounds like future work.

Can we agree for now we'll focus exclusively on #1036 and keep close this PR and #1007 ?

uvjustin:
> OK thanks for the clarification regarding sample_rate. It seems to me we'll at some point need to make up our minds whether we should be providing direct access to `codecpar` or to the codec context, but that sounds like future work.
> 
> Can we agree for now we'll focus exclusively on #1036 and keep close this PR and #1007 ?

Sounds good to me

PR #1035, opened 2022-10-17
Clarify "rate" properties by deprecating aliases

jlaine:
We have multiple properties referring to "rates", some of which go by a generic "rate" name without specifying whether it is a frame rate or a sample rate.

This change deprecates:

- `AudioCodecContext.rate` (use `AudioCodecContext.sample_rate`)
- `AudioFrame.rate` (use `AudioFrame.sample_rate`)
- `VideoCodecContext.rate` (use `VideoCodecContext.framerate`)

### DIFF
--- a/av/audio/codeccontext.pyx
+++ b/av/audio/codeccontext.pyx
@@ -1,5 +1,6 @@
 cimport libav as lib
 
+from av import deprecation
 from av.audio.format cimport AudioFormat, get_audio_format
 from av.audio.frame cimport AudioFrame, alloc_audio_frame
 from av.audio.layout cimport AudioLayout, get_audio_layout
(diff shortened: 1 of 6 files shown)


jlaine:
I'm not 100% sure we actually want this PR.

uvjustin:
> I'm not 100% sure we actually want this PR.

Yeah, it might not be necessary. rate seems like the most canonical name for referring to the sample_rate/framerate. The main motivation for #1034 was actually from your comment here: https://github.com/PyAV-Org/PyAV/pull/1007#issuecomment-1187648549 suggesting that we didn't need/want these aliases.

jlaine:
OK I'm going to close this.

PR #1036, opened 2022-10-17
Deprecate Stream.framerate passthrough

uvjustin:
See https://github.com/PyAV-Org/PyAV/issues/1005
Since the behavior of `VideoStream.framerate` has changed, this PR disallows passthrough of that attribute so that users will be forced to switch to `VideoStream.average_rate`

### DIFF
--- a/av/stream.pyx
+++ b/av/stream.pyx
@@ -1,3 +1,5 @@
+import warnings
+
 from cpython cimport PyWeakref_NewRef
 from libc.stdint cimport int64_t, uint8_t
 from libc.string cimport memcpy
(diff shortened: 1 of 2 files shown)


jlaine:
I think I'd rather raise a warning here with more detailed information about why we don't recommend using `framerate`. We also need to catch access to `rate` *if it's a video stream*. I don't think we need to worry about the setter.

jlaine:
@uvjustin I have amended the commit to use an explicit deprecation warning, would you be satisfied with this?

uvjustin:
> @uvjustin I have amended the commit to:
> 
> * handle both `framerate` and `rate`
> * use an explicit deprecation warning including the rationale for the change
> 
> Would you be satisfied with this?

That works for me. A possibly simpler alternative may be to just use the deprecation helper in `VideoStream` directly:
```
cdef class VideoStream(Stream):

    rate = deprecation.renamed_attr("average_rate")
    framerate = deprecation.renamed_attr("average_rate")
```

Issue #1060, opened 2022-12-24
Exposing frame buffer array


hmaarrfk:
Would you consider exposing the frame buffer array either by exposing `av_image_fill_pointers` instead of always allocating an array, or by making it available for writing as a numpy array?

I understand that you would like to align the buffer to 16 (bytes???) but some applications may have already worked to ensure that their data is aligned to better than that.

Finally, it would be good if some precessing could share that buffer and make it an "output" buffer.

Just a thought. It might help our high performance applications.

https://github.com/PyAV-Org/PyAV/blob/bc4eedd5fc474e0f25b22102b2771fe5a42bb1c7/av/video/frame.pyx#L95-L106

hmaarrfk:
Not stale

hmaarrfk:
not stale.

hmaarrfk:
.

hmaarrfk:
Thanks!

PR #1061, opened 2022-12-24
Create a path toward sharing memory from a numpy buffer

hmaarrfk:
I know there is some concern about having the memory aligned.

I didn't know if it needed to be aligned to 16 bits, or 16 bytes.

If you know let me know which it is, I can add a check for it. If you don't want to support this, then that is fine too.

Closes: https://github.com/PyAV-Org/PyAV/issues/1060

### DIFF
--- a/av/video/frame.pxd
+++ b/av/video/frame.pxd
@@ -11,6 +11,7 @@ cdef class VideoFrame(Frame):
     # This is the buffer that is used to back everything in the AVFrame.
     # We don't ever actually access it directly.
     cdef uint8_t *_buffer
+    cdef object _np_buffer
 
     cdef VideoReformatter reformatter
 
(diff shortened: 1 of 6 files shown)


hmaarrfk:
@jlaine I didn't ask but are you interested in something like this? I've been using it with quite a bit of success more recently, I can add checks for continuous buffers too if you want.

hmaarrfk:
anycomments on the desire to even support this kind of feature?

hmaarrfk:
I've rebased this. this is ready for review.

WyattBlue:
For the future, I'm okay with having numpy as a compile time dep if that meant PyAV was faster or smaller.

hmaarrfk:
> For the future, I'm okay with having numpy as a compile time dep if that meant PyAV was faster or smaller.

Maybe we can discuss this in an issue

PR #1067, opened 2023-01-02
Add the flag ENCODER_FLUSH

hmaarrfk:
Available since FFMPEG 4.3

I can help cleanup the flag names in a subsequent PR. it seems that you "support FFMPEG >= 4.3")


hmaarrfk:
I'm somewhat "confirmed" that this has some effect.

* libopenh264 reports false
* nvenc reports true

at least something is happening.

jlaine:
Could you please rebase on top of `main`, CI should be fixed.

Issue #1045, opened 2022-10-25
v10 no longer writes "rotate" metadata flag


FirefoxMetzger:
## Overview

As of v10.0.0 the stream metadata flag `rotate` is no longer written to file. This may affect other metadata fields; however, for this one specifically, we have a unit-test downstream so I know for sure that it is affected.

## Expected behavior

Setting a field in a video stream's metadata dict causes it to be written to the file.

## Actual behavior

Stream-level metadata (at least the `rotation` flag) is no longer written.

## Reproduction

```python
import av
import numpy as np

print(f"PyAV version: {av.__version__}")

with av.open("test.mp4", "w") as container:
    container.metadata["comment"] = "This video has a rotation flag."

    stream = container.add_stream("libx264", 24)
    stream.metadata["rotate"] = "90"

    for _ in range(24):
        frame = av.VideoFrame.from_ndarray(np.ones((128, 128, 3), dtype=np.uint8))
        for packet in stream.encode(frame):
            container.mux(packet)

    for packet in stream.encode():
        container.mux(packet)

with av.open("test.mp4") as container:
    print(container.metadata)
    video_stream = container.streams.video[0]
    print(video_stream.metadata)

```

Using av v9.2 the above snippet produces:
```
PyAV version: 9.2.0
{'major_brand': 'isom', 'minor_version': '512', 'compatible_brands': 'isomiso2avc1mp41', 'encoder': 'Lavf58.76.100', 'comment': 'This video has a rotation flag.'}
{'rotate': '90', 'language': 'und', 'handler_name': 'VideoHandler', 'vendor_id': '[0][0][0][0]'}
```

However, when switching to v10 it produces:
```
PyAV version: 10.0.0
{'major_brand': 'isom', 'minor_version': '512', 'compatible_brands': 'isomiso2avc1mp41', 'encoder': 'Lavf59.27.100', 'comment': 'This video has a rotation flag.'}
{'language': 'und', 'handler_name': 'VideoHandler', 'vendor_id': '[0][0][0][0]'}
```

## Versions

- OS: Windows 11 and Ubuntu 20.04
- PyAV runtime:

```
PyAV v10.0.0
[...]

PR #1062, opened 2022-12-25
[streams] populate added stream with codec param (fixes #1044)

HanzCEO:
fixes #1044 

avformat_new_stream(AVFormatContext *s, const AVCodec *c) does not use their second parameter [[1]]
So avcodec_parameters_from_context() should do the job [[2]]

[1]: https://ffmpeg.org/doxygen/trunk/group__lavf__core.html#gadcb0fd3e507d9b58fe78f61f8ad39827
[2]: https://ffmpeg.org/doxygen/trunk/codec__par_8c_source.html#l00099

PR #1068, opened 2023-01-03
BUGFIX: fix segfault on accessing stream type prior to any encoding

hmaarrfk:


jlaine:
Looks like a good catch, can you please add a test exercising this condition?

hmaarrfk:
Hopefully the test is clear. I hope I got the test directory correct.

tests/test_streams.py
```diff
@@ -27,3 +28,21 @@ def test_selection(self):
         self.assertEqual([video], container.streams.get(video=(0,)))
 
         # TODO: Find something in the fate suite with video, audio, and subtitles.
+
+    def test_stream_properties(self):
+        # Ensure that all stream properties have sensible values even
+        # if writing to the stream hasn't started yet
+        with av.open(self.sandboxed("output.mp4"), "w") as container:
+            stream = container.add_stream('h264')
```
jlaine, line 36: Why is the type unknown when *we* add a stream?

hmaarrfk: It could be. I believe that it is because the codec context is only "opened" on first use.

TBH, I only noticedthis since accessing the `__repr__` when running my code interactively cause things to crash.

hmaarrfk: You are correct, with #1062 (and my "fix" removed) applied I should update this test with:

```
diff --git a/tests/test_streams.py b/tests/test_streams.py
index efee029..496d32d 100644
--- a/tests/test_streams.py
+++ b/tests/test_streams.py
@@ -45,4 +45,4 @@ class TestStreams(TestCase):
             assert stream.duration is None
             assert stream.frames == 0
             assert stream.language is None
-            assert stream.type is None
+            assert stream.type == 'video'
```

jlaine:
PR #1062 is going to land if CI doesn't blow up. This still leaves us with the valid segfault you've spotted, but now I'm not sure how to test it anymore..

hmaarrfk:
I'm not sure it is so valid anymore. If the pointer is guaranteed to exist, then it is likely that the code is ok right?

I

hmaarrfk:
Selfishly, i would much rather you spend your precious time helping us resolve https://github.com/PyAV-Org/PyAV/issues/1045

PR #1074, opened 2023-01-11
Allow setting the max_bit_rate and min_bit_rate in codec context

hmaarrfk:
These variables are documented as part of ffmpeg.

It might be useful for others to read and write them.

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -593,10 +593,17 @@ cdef class CodecContext(object):
 
     property max_bit_rate:
         def __get__(self):
-            if self.ptr.rc_max_rate > 0:
-                return self.ptr.rc_max_rate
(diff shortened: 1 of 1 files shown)


jlaine:
OK could we have a test for this please?

hmaarrfk:
I'm actually no longer sure *what* should happen when these things change. I had to specificy the codec context options dictionary to have a visible change in my output quality.

jlaine:
I'm guessing this PR hasn't been rebased yet as the new tests are against `pypy3.9` and not `pypy3`?

hmaarrfk:
I've rebased the patch, but I'm not sure what effect "writing" to these two properties has.

WyattBlue:
This PR is too old.

PR #1077, opened 2023-01-21
[scratchpad] xrange() was removed in Python 3

cclauss:
https://docs.python.org/3/whatsnew/3.0.html#views-and-iterators-instead-of-lists
> [`range()`](https://docs.python.org/3/library/stdtypes.html#range) now behaves like `xrange()` used to behave, except it works with values of arbitrary size. The latter no longer exists.

### DIFF
--- a/scratchpad/audio.py
+++ b/scratchpad/audio.py
@@ -13,7 +13,7 @@ def print_data(frame):
         data = bytes(plane)
         print('\tPLANE %d, %d bytes' % (i, len(data)))
         data = data.encode('hex')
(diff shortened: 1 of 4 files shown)


jlaine:
Thanks for catching this. FYI I believe the "scratchpad" was an area where Mike put examples which still needed to be polished so I wouldn't be surprised if it has fallen out of date.

jlaine:
The CI issue should be solved by #1040 would you mind rebasing?

jlaine:
I'm going to close this since #1079 has both your commits

PR #1079, opened 2023-01-23
[scratchpad] buffer() was removed in Python 3

cclauss:
As discussed at https://portingguide.readthedocs.io/en/latest/etc.html#raw-buffer-protocol-buffer-and-memoryview `buffer()` was removed in Python 3 in favor of [`memoryview()`](https://docs.python.org/3/library/stdtypes.html#memoryview).

### DIFF
--- a/scratchpad/cctx_encode.py
+++ b/scratchpad/cctx_encode.py
@@ -1,14 +1,13 @@
 import logging
 
-from PIL import Image, ImageFont, ImageDraw
+from PIL import Image, ImageDraw, ImageFont
 
-logging.basicConfig()
-
-import av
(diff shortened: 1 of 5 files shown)


scratchpad/cctx_encode.py
```diff
@@ -35,7 +34,7 @@
     packet = cc.encode(frame)
     print('   ', packet)
 
-    fh.write(str(buffer(packet)))
+    fh.write(str(memoryview(packet)))
```
jlaine, line 37: This actually looks wrong, the file handle should have been opened with `open('test.flv', 'wb')` and then writing is just `fh.write(bytes(packet))`.

PR #1099, opened 2023-03-09
Update README.md

brentpicasso:
Enhance instructions for building when no binary packages are available.


jlaine:
Hi! This looks like it's redundant?

https://github.com/PyAV-Org/PyAV/blob/main/docs/overview/installation.rst

PR #1108, opened 2023-03-21
fix the doc about `time`

zzjjbb:
Fix the doc of the `time` page and `av.frame.Frame.dts` according to https://ffmpeg.org/doxygen/5.1/structAVFrame.html#aa52951f35ec9e303d3dfeb4b3e44248a

Also, is it worth to change `av.frame.Frame.dts` to `av.frame.Frame.pkt_dts` to avoid confusion?

### DIFF
--- a/av/frame.pyx
+++ b/av/frame.pyx
@@ -68,7 +68,9 @@ cdef class Frame(object):
 
     property dts:
         """
-        The decoding timestamp in :attr:`time_base` units for this frame.
+        The decoding timestamp copied from the :class:`~av.packet.Packet` that triggered returning this frame in :attr:`time_base` units.
+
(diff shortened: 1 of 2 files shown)


av/frame.pyx
```diff
@@ -68,7 +68,9 @@ cdef class Frame(object):
 
     property dts:
         """
-        The decoding timestamp in :attr:`time_base` units for this frame.
+        The decoding timestamp copied from the :class:`~av.packet.Packet` that triggered returning this frame in :attr:`time_base` units.
```
jlaine, line 71: Is there a link to the upstream documentation we could include?

zzjjbb: I've included the link in the PR description: https://ffmpeg.org/doxygen/5.1/structAVFrame.html#aa52951f35ec9e303d3dfeb4b3e44248a

PR #1119, opened 2023-04-12
Add max_b_frames property to codec context

davidplowman:
Allows an application to control the maximum number of B frames, irrespective of what other settings or preset is used.

### DIFF
--- a/av/video/codeccontext.pyx
+++ b/av/video/codeccontext.pyx
@@ -161,3 +161,15 @@ cdef class VideoCodecContext(CodecContext):
     property coded_height:
         def __get__(self):
             return self.ptr.coded_height
+
+    property max_b_frames:
+        """
+	The maximum run of consecutive B frames when encoding a video.
+
+        :type: int
+        """
+        def __get__(self):
+            return self.ptr.max_b_frames
+
+        def __set__(self, value):
+            self.ptr.max_b_frames = value
--- a/tests/test_encode.py
+++ b/tests/test_encode.py
@@ -1,6 +1,9 @@
 from fractions import Fraction
 from unittest import SkipTest
+import io
 import math
+
+import numpy as np
 
 from av import AudioFrame, VideoFrame
 from av.audio.stream import AudioStream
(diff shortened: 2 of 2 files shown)


davidplowman:
Hi again, just wondering if there is anything else I need to do to get this PR considered? Obviously it would be really nice for us if we can just tell our users to install the standard pyav package. Thanks!

av/video/codeccontext.pyx
```diff
@@ -162,3 +162,10 @@ cdef class VideoCodecContext(CodecContext):
     property coded_height:
         def __get__(self):
             return self.ptr.coded_height
+
+    property max_b_frames:
+        def __get__(self):
```
jlaine, line 167: Could we have a docstring explaining the use of this property please?

Does it make sense for decoding / encoding / both?

jlaine:
We have a policy of requiring unit tests for new code. Depending on whether this flag makes sense for encoding or decoding you'll  most likely want to add some test code in `test_encode.py` or `test_decode.py`.

davidplowman:
Hi, thanks for your reply. I'm a bit busy with other things at the moment but will get back to this just as soon as I can. Would you prefer the test code in a 2nd commit in this PR, or all merged together in one commit?

We're using PyAV as the interface from Python to libav on the new Raspberry Pi 5 now, so I'm very keen to try and un-fork the version that we're currently distributing. Thanks!

davidplowman:
The latest update adds a docstring and a test for the functionality. Please let me know if it needs re-formatting or changing. Thanks!

davidplowman:
Thanks for running the tests. Obviously I'm not experienced at understanding these logs, but looking at the failure, I'm suspicious that this is just a timeout, and that nothing is actually "wrong". What do you think? Am I right in thinking the ubuntu aarch64 platform is much slower than the others - do you happen to know what kind of machine it is? Thanks!

WyattBlue:
The one failing test is just aach64 timing out (not the fault of this PR). Everything looks good so I'm going to merge this.

PR #1133, opened 2023-06-18
Enable or disable logging

eerimoq:
The current log callback sometimes deadlocks when I use aiortc with [this line](https://github.com/aiortc/aiortc/blob/main/src/aiortc/__init__.py#L53) commented out, which uses this package. I couldn't figure out how to properly fix it, so I implemented a workaround that allows the user to totally enable and disable FFmpeg logging.

### DIFF
--- a/av/logging.pyx
+++ b/av/logging.pyx
@@ -121,6 +121,15 @@ def restore_default_callback():
     lib.av_log_set_callback(lib.av_log_default_callback)
 
 
+cdef bint enabled = True
+
+
+def set_enabled(value):
+    """Enable or disable all logging."""
+    global enabled
+    enabled = value
+
+
 cdef bint print_after_shutdown = False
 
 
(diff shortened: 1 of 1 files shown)


jlaine:
I'm not sure this is an actual fix, it adds more complexity to the logging situation with yet another flag. Have you considered any alternatives?

eerimoq:
I don't think I did. Feel free to just close this PR as it was mostly meant to pitch the idea.

jlaine:
To be clear, as mentioned in #1186 I think we absolutely do need to revisit the logging framework. Unless someone finds a way to fix the deadlocks introduced by logging calling back into Python I'd be in favour of ripping it out, or at least having it off by default. 

.. so let's keep the PR Alive as a starting point!

hmaarrfk:
It seems to be deadlocking due to the GIL. 

are you in a heavy multi-threading environment?

Could you provide some reproducing code that deadlocks even 10% of the time?

I would like to explore solutions where a C++ queue is made, and flused upon request maybe....

eerimoq:
Sorry, this was such a long time ago I barely remember what I did.

hmaarrfk:
understood. lets leave this open as an acknowledgement of the issue.

For future contributors, to start taking action on this front, it would be great to have some code that could reproduce even 10% of the time.

When submitting the reproducer, try too include information about the processor used in the system and the number of threads it supports.

WyattBlue:
Superseded by #1383

PR #1138, opened 2023-06-27
expose discard, trusted and disposable packet flags

skirklin:
I have a use case where the video I am parsing has "dummy" packets inserted, marked with AV_PKT_FLAG_DISPOSABLE and intended to be skipped over when decoding and found that I needed to extend the Packet class to be able to detect it. I do not know if there are any compatibility concerns, i.e. if these flags are a part of all supported versions of avcodec.h but wanted to at least offer the patch in case it might be helpful for someone else out there.


lletourn:
not too sure what's the process of voting in PyAV, but I too would like this functionality merged in.

WyattBlue:
Continued in #1327

Issue #1140, opened 2023-07-18
Pip install is failing to build due to av/logging.pyx error


nicholas-presien:
## Overview

I'm currently using the stable version of PyAV (v9.0.2) in a docker container, and building in a docker image for a while with no issue. Recently, my docker image failed due to build errors in pyav when installing pyav using pip.

I'm building on a TX2NX dev kit, running Jetpack 4.6 with an arm64 compiled version of FFMPEG 4.3.2(https://launchpad.net/~jonathonf/+archive/ubuntu/ffmpeg-4), on python v3.6.9 

## Expected behavior

It should build without error

## Actual behavior

It's now building with error, when nothing in the environment has changed due to it being in a docker container.

Build report:
```
python3 setup.py build_ext --inplace                                                                                                                                                                           
Compiling av/logging.pyx because it changed.                                                                                                                                                                                                  
[1/1] Cythonizing av/logging.pyx                                                                                                                                                                                                              
                                                                                                                                                                                                                                              
Error compiling Cython file:                                                                                                                                                                                                                  
------------------------------------------------------------                                                                                                                                                                                  
...                                                                                                                                                                                                                                           
cdef const char *log_context_name(void *ptr) nogil:                                                                                                                                                                                           
    cdef log_context *obj = <log_context*>ptr                                                                                                                                                                                                 
    return obj.name                                                                                                                                                                                                                           
                                                                                                                                                                                                                                              
cdef lib.AVClass log_class                                                                                                                                                                                                                    
log_class.item_name = log_context_name                                                                                                                                                                                                        
                      ^                                                                                                                                                                                                                       
------------------------------------------------------------                                                                                                                                                                                  
                                                                                                                                                                                                                                              
av/logging.pyx:216:22: Cannot assign type 'const char *(void *) except? NULL nogil' to 'const char *(*)(void *) noexcept nogil'                                                                                                               
                                                                                                                                                                                                                                              
Error compiling Cython file:                                                                                                                                                                                                                  
------------------------------------------------------------                                                                                                                                                                                  
...                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                              
# Start the magic!                                                                                                                                                                                                                            
# We allow the user to fully disable the logging system as it will not play                                                                                                                                                                   
# nicely with subinterpreters due to FFmpeg-created threads.                                                                                                                                                                                  
if os.environ.get('PYAV_LOGGING') != 'off':                                                                                                                                                                                                   
[...]

PR #1145, opened 2023-07-23
Small updates to enable Cython 3

hmaarrfk:
Mostly it seems that cython wants you to declare things as `noexcept`

Closes https://github.com/PyAV-Org/PyAV/issues/1140

I'm going to see if I can report the bug to cython, but it is hard for me to create a minimum reproducible example


hmaarrfk:
Tested locally and cython 0.29.36 and it seems to compile and function just as well.

av/video/format.pyx
```diff
@@ -47,10 +47,14 @@ cdef class VideoFormat(object):
         self.ptr = lib.av_pix_fmt_desc_get(pix_fmt)
         self.width = width
         self.height = height
-        self.components = tuple(
+        # hmaarrfk -- 2023/07/23
+        # Note on tuple([])
+        # Cython 3 seems to have trouble with cdef tuples, so we use a list
+        # it complains about some const identifier not being able to get assigned
```
SpecLad, line 53: FWIW, this issue was fixed in Cython 3.0.1, so this workaround can probably be removed.

moi90:
It works for me locally, too. I hope this gets merged soon!

rvanlaar:
What's needed to get this merged?

jlaine:
> What's needed to get this merged?

Co-maintainers!

av/video/format.pyx
jlaine: If this isn't needed can we prune it please?

docs/development/includes.py
```diff
@@ -107,9 +107,16 @@ def extract(path, **kwargs):
         c_string_encoding='ascii',
     )
 
-    context = options.create_context()
+    context = Context(
```
jlaine, line 110: How do these changes relate to the PR?

jlaine:
Can you point us to some Cython docs explaining why these changes are required, and if there are others we need to look out for?

docs/development/includes.py
hmaarrfk: cython3 removed create_context, not sure when so I looked at the old implementation of create_context and copied it here.

hmaarrfk:
I'm not sure there is much to point to:

1. Cython wanted to optimize more so they ask that you write `noexcept`. Search for `noexcept` in https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html
2. There was a bug that I reported upstream https://github.com/cython/cython/issues/5558 that got fixed.
3. I'm not sure how I found the `create_context` was removed. sometimes I dig deep into code. I don't have a good reference for how I did this months ago

SpecLad:
> Can you point us to some Cython docs explaining why these changes are required, and if there are others we need to look out for?

It's explained here: https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept

hmaarrfk:
i'm not sure why the tests are failing.

docs/development/includes.py
ENate: Do we have to add this in the dev folder? There no such folder available inside transformers

PR #1195, opened 2023-10-28
Update CI/CL, drop Python 3.7

WyattBlue:
Python 3.7 is EOL, let's at least use 3.8 as the minimum.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,15 +1,19 @@
 name: tests
-
-on: [push, pull_request]
-
+on:
+  push:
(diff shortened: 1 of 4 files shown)


WyattBlue:
Merging since the failing tests are complaining about Cython code, something out of scope for this commit.

jlaine:
I'd rather we didn't merge PRs when CI is failing. If we have flakey tests.. fix them!

WyattBlue:
I'd rather we "fail-forward" instead of making one mega-commit that fixes everything

PR #1197, opened 2023-10-28
Fix CI/CL

WyattBlue:


jlaine:
Some points:
- one commit = one change, I'm not fond of "mixed bag" commits which change a whole bunch of things
- commented code is a code smell: either we want the code or we don't (we have the git history).
-  I haven't looked into the detail but why comment out a failing test? It used to pass, if it doesn't it's a sign either the test needs redesigning or something has changed under our feet

WyattBlue:
@jlaine Your points are correct. I'm just trying to figure out why the tests are failing, then I'm clean up the PR.

hmaarrfk:
I came here because I was curious to see the progress. 

I learned that it is possible, if you want to have large style changes as this one, to ignore a specific commit in the git blame.

jlaine:
> @jlaine Your points are correct. I'm just trying to figure out why the tests are failing, then I'm clean up the PR.

Sounds good! Again kudos for the energy you're putting into this, don't get me wrong I'm very grateful for your involvement :)

jlaine:
AFAICT the test that was "commented" is in fact `test_doctests.py` which is completely removed? Could we have the rationale for this and have it in a separate commit please?

What tool did you use to reformat the tests? Currently we run `black`, `flake8` and `isort` (the two last ones might be replaced by `ruff` if it supports Cython code).

WyattBlue:
black is not being applied to the `tests` directory.

PR #1199, opened 2023-10-29
Fix running doctests

jlaine:
We use `av.datasets` so explicitly import it

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,10 +57,8 @@ jobs:
         config:
           - {os: ubuntu-latest, python: 3.8, ffmpeg: "5.1", extras: true}
           - {os: ubuntu-latest, python: 3.8, ffmpeg: "5.0"}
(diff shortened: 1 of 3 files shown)


jlaine:
@WyattBlue it looks as though the changes for FFmpeg 6 dropped compatibility with FFmpeg < 5.0. This is fine by me but then let's make 5.0 our baseline including in docs:

https://github.com/PyAV-Org/PyAV/blob/main/docs/overview/installation.rst?plain=1#L59

.. and the CI matrix.

WyattBlue:
@jlaine Got it. Let's make https://github.com/PyAV-Org/PyAV/pull/1198 where we apply these changes.

bedilbek:
@jlaine  I was just curious why have you decided to drop support for FFmpeg 4.x version, maybe at least support for 4.4 could be there for some more time

WyattBlue:
@bedilbek We're dropping support for ffmpeg 5 full stop.

PR #1203, opened 2023-10-30
Expose bits_per_coded_sample on VideoCodecContext

rvanlaar:
qtrle decoder needs `bits_per_coded_sample` to be set when decoding a packet.


jlaine:
Can we have a unit test which exercises the getter and setter please?

av/video/codeccontext.pyx
```diff
@@ -92,6 +92,14 @@ cdef class VideoCodecContext(CodecContext):
             self.ptr.height = value
             self._build_format()
 
+    property bits_per_coded_sample:
+        def __get__(self):
```
jlaine, line 96: This needs a docstring

av/video/codeccontext.pyx
```diff
@@ -92,6 +92,14 @@ cdef class VideoCodecContext(CodecContext):
...
+            return self.ptr.bits_per_coded_sample
+
+        def __set__(self, unsigned int value):
```
jlaine, line 99: Why an *unsigned* int?

WyattBlue:
In addition to jlaine's comments, this needs to be redone with the new style property attribute and it needs the type hint.

PR #1210, opened 2023-11-01
Add codec parameters

daveisfera:


WyattBlue:
How does this PR improve PyAV?

daveisfera:
Other than allowing `PyAV` to more closely/accurately match the underlying `ffmpeg` definitions, it allows for specifying values to `add_stream`
I'm creating a stream without an input stream that I can pass as `template` and it currently runs into problems because the method only allows setting a small number of values on the stream itself rather than the context. Basically, fleshing out the rest of the values for `AVCodecParameters` was my real need and the rest of the changes were just clean up while I was in there

daveisfera:
I was able to fix this problem without this change

jlaine:
Can you maybe share your solution?

daveisfera:
It appears to be at least slightly related to #1225 but when you do `output.add_stream("hevc")` it incorrectly converts it to `libx265` and then several parts of the processing don't know what to do. If instead you do `output.add_stream(Codec("hevc", "r"))` then you can end up with other warnings about writing to a decoder, but it at least works. There's probably a "real fix" here, but I wasn't able to figure out where the breakdown was happening and this workaround got it working, so I'm running with it

PR #1211, opened 2023-11-01
Cleanup

daveisfera:


jlaine:
Can you please set aside the changes to "scratchpad" I think I'll just drop that directory. For the rest of the style changes, could we enforce this in CI please?

daveisfera:
Done and I added the `ruff` check, but it will fail until the `scratchpad` stuff is removed and some other remaining issues are cleaned up

jlaine:
Instead of depending on yet another third-party action, could you please just add ruff to:

- The test requirements: https://github.com/PyAV-Org/PyAV/blob/main/tests/requirements.txt
- The `lint` target in `Makefile` https://github.com/PyAV-Org/PyAV/blob/de5a3b365fe6e0f5dd7398f1c34a4bfb736d5f4b/Makefile#L27

That way it will get run as part of our "style" job, and we can explicitly state which directories to lint and adjust `ruff` configuration in `pyproject.toml` if needed.

Also can we flesh out the commit messages a bit, as they are extremely terse and don't include any motivation for the changes. For instance why the change to `AVError`?

jlaine:
EDIT: in fact running `ruff` seems to be beside the point, it's not picking up any additional issues. Since this PR is focusing on Cython code, it should do only that.

WyattBlue:
I removed all the unrelated changes, so it's good to go.

PR #1214, opened 2023-11-02
Add action to check code with ruff

daveisfera:
Adding an [action to check the code with `ruff`](https://github.com/PyAV-Org/PyAV/pull/1211#issuecomment-1789599218)

I made it from the main branch, because I wanted to verify that it would identify issues when they exist


daveisfera:
Just wanted to verify that the `ruff` check would fail when there were issues and I added the real thing in #1211

PR #1222, opened 2023-11-07
Improve actions

WyattBlue:


jlaine:
Let's have more descriptive commit messages. The git history is all we'll have months or years from now to understand why a change was made. How are the actions "improved"?

PR #1230, opened 2023-11-16
Prevent duplicate test runs on new PRs

WyattBlue:


jlaine:
As previously commented I don't see why this is necessary, can you show me an instance of duplicate runs?

WyattBlue:
An example:
Screenshot 2023-11-16 at 2 15 39 PM
One push, one merge, yet 3 CI runs. There would only be 2 if this commit was applied.

PR #1231, opened 2023-11-20
Allow using pathlib.Path for av.open

WyattBlue:


hmaarrfk:
upcasting int to float is typically an "ok" thing to do.

hmaarrfk (approved):

PR #1248, opened 2023-12-17
Acknowledge recent changes in docs

WyattBlue:
Update documentation to what we do now. Point to active url


WyattBlue:
For context, basswood-io.com is my domain for my company. I think we should go with these changes since I don't have access to the original pyav.org domain.

Issue #1150, opened 2023-08-08
Feature request: Add side_data attribute to Stream


hyenal:
## Overview

Sometimes a stream encodes side data in the entire stream (in [side_data](https://ffmpeg.org/doxygen/trunk/structAVStream.html)). Some of those side data are quite essential to read the video correctly.
For instance `displaymatrix` sometimes encode whether the video is rotated or not.

From the AV documentation it seems like we could know easily if there is side data with the attribute `nb_side_data`.

I would like to know if this is an interesting feature for this library ? I started a fork [there](https://github.com/hyenal/PyAV/tree/add-stream-side-data) with, for now, a simple skeleton of how it would look like

## Desired Behavior

When reading a video (or any other stream) you would get the side data as a dictionary:
```python
# This would give you the stream metadata
container.stream.video[0].side_data
```

hyenal:
I made [this PR](https://github.com/hyenal/PyAV/pull/1/files) on my own fork. It only fetch the DISPLAYMATRIX but would be happy to implement more and submit a PR to this repo if anyone is interested

lgeiger:
@hyenal I'd be really interested in having this functionality upstreamed to `av` as it would fix #1045 is a blocker for me currently and also many people seem to run into.

@jlaine or @WyattBlue are probably in a better position to judge how this should be implemented best to make it consistent with the way frame side data is handled. I'm happy to help out and write some test cases if that would be helpful.

hyenal:
Thanks @lgeiger, so happy to see this getting momentum!

It looks like it's been merged in @WyattBlue repository and will be available in PyAv soon? 

I am also happy to write some tests if this helps :)

lgeiger:
> It looks like it's been merged in @WyattBlue repository and will be available in PyAv soon?

Yes, it's already available on [pypi](https://pypi.org/project/pyav) though this causes dependency conflicts for me as I am relying on a different package that requires `av` directly which is why it would be great to get upstreamed at some point.

hyenal:
> Yes, it's already available on [pypi](https://pypi.org/project/pyav) though this causes dependency conflicts for me as I am relying on a different package that requires `av` directly which is why it would be great to get upstreamed at some point.

I see thanks for the explanation!

Let's wait for an answer from maintainers then, I am hopeful this could be merged 🙏

WyattBlue:
Sure, @hyenal Make a new PR for this repo.

hyenal:
[...]

PR #1249, opened 2023-12-19
Expose DISPLAYMATRIX side data

hyenal:
This PR allows users to get DISPLAYMATRIX side data as part of the video stream attribute

It solves [this issue](https://github.com/PyAV-Org/PyAV/issues/1150).

While `SideData` is only populated with `DISPLAYMATRIX` so far, one could easily add other side data in future PR and query them via `_get_side_data`.

All tests are passing locally


WyattBlue:
Looks good. I don't it makes sense to put a docstring on a internal method so removed it. I also removed using the SideData object in cdef code because it's slower than comparing the int directly.

PR #1252, opened 2023-12-27
Upgrade to ffmpeg 6.1

WyattBlue:


jlaine:
This PR does not do what it says: it is also dropping all tests against FFmpeg 5.0, why?

jlaine:
DlPlease do NOT merge until you have answered my question.

WyattBlue:
> it is also dropping all tests against FFmpeg 5.0, why?
1. Actions Minutes are not unlimited
2. We don't need to test every minor release. ffmpeg is good at following semvar and we already should be relying on obscure implementation details
3. ffmpeg 7.0 is coming out soon, and I assume when that happens, our support for the 5.x series will be dropped

WyattBlue:
> This PR does not do what it says

This is an odd comment. This commit tests with 6.1 and will ship the ffmpeg 6.1 binaries instead of the 6.0 binaries when it's released on PyPI.
6.1 does set average_rate to `25/1` instead of null on empty streams now when it didn't before so I removed that assert. This might not be the best way to handle it, but I wanted to make sure nothing else was breaking.

jlaine:
> > This PR does not do what it says
> 
> This is an odd comment. 

My point is that PRs and commit messages need to actually describe what they are doing: this one adds tests against FFmpeg 6.1 (good), but also drops tests against FFmpeg 5.0 - which we say we still support (not good). Either we still support FFmpeg 5.0, in which case we keep testing against it or we officially abandon support FFmpeg 5.0.

I'm not adverse to dropping FFmpeg 5.0 but this needs to be reflected both in the commit message and the documentation (`docs/overview/installation.rst:`).

FFmpeg minor versions can also introduce new APIs or deprecate old ones: I think this is exactly what happens with FFmpeg 5.1, I noticed we're getting compilation warnings specifically in FFmpeg 5.1 and up `‘channel_layout’ is deprecated`.. but the replacement API doesn't exist in FFmpeg 5.0. So if we fix this deprecation warning we will be breaking FFmpeg 5.0 compatibility, but CI won't tell us that because we have stopped testing against 5.0!

WyattBlue:
Okay, I've restored all ffmpeg 5.0 and 5.1 tests.

Issue #1254, opened 2024-01-07
Incorrect value returned by VideoCodecContext.gop_size


JoeUgly:
## Overview
It is my understanding that "GOP size" refers to the number of frames between each keyframe.
However, the value returned by `VideoCodecContext.gop_size` differs from what I get when looking at the index of each keyframe.

## Expected behavior
All methods of determining the GOP size or keyframe intervals should yield the same results.

## Actual behavior
For all the videos I tested `VideoCodecContext.gop_size` returns 12. Other methods indicate different GOP sizes (such as 30 or 120 depending on the media).

## Investigation
The ffmpeg documentation seems to indicate that this property is used only for setting GOP size when encoding:
> decoding: unused

[https://ffmpeg.org/doxygen/trunk/structAVCodecContext.html#a9b6b3f1fcbdcc2ad9f4dbb4370496e38](https://ffmpeg.org/doxygen/trunk/structAVCodecContext.html#a9b6b3f1fcbdcc2ad9f4dbb4370496e38)

I'm not sure where the value returned by `VideoCodecContext.gop_size` comes from. It might be from the libavcodec generic global options:
> Set the group of picture (GOP) size. Default value is 12.

[https://ffmpeg.org/ffmpeg-codecs.html](https://ffmpeg.org/ffmpeg-codecs.html)

## Reproduction
The following code shows 2 ways of getting the GOP size:
```
import av

path = "/home/joepers/code/test_pyav/short_1080.webm"

keyframe_l = []

with av.open(path) as container:
    stream = container.streams.video[0]
    stream.thread_type = "AUTO"
    print(f"gop_size: {stream.codec_context.gop_size}")

    for frame in container.decode(stream):
        if frame.key_frame:
            keyframe_l.append(frame.index)


print(f"keyframe indices: {keyframe_l}")  # Manually inspect keyframe interval
unique_differences = set([j-i for i, j in zip(keyframe_l[:-1], keyframe_l[1:])])  # Auto inspect keyframe interval
print(f"number of frames between keyframes: {unique_differences}")
```
`stream.codec_context.gop_size` should be the same as the distance between each keyframe index.

## Versions

- OS: Ubuntu 23.10
- PyAV runtime:
```
PyAV v11.0.0
[...]

PR #1255, opened 2024-01-07
removed unsupported VideoCodecContext.gop_size.__get__ method

JoeUgly:
Fix [this issue](https://github.com/PyAV-Org/PyAV/issues/1254).

I believe ffmpeg does not support getting the GOP size, therefore the returned value is incorrect.

This is my first pull request so let me know if I'm doing it wrong.


jlaine:
I don't think this is the right approach, see my comments in #1254

Issue #1158, opened 2023-08-29
av.VideoFrame index not updated properly when "thread_type" is set to AUTO.


swillisart:
## Overview

If I **demux** or **decode** a video stream using thread_type = 'AUTO', produced frames at the tail end of the stream get incorrect indices. (These will change based on the systems thread count)

If it set the thread_type = 'NONE' it will correctly produce the frame number indices.

## Expected behavior

av.VideoFrames generating indices representing a match to their frame number.

## Actual behavior
Given any video (I do not think it's codec specific happens on both x264 and MPEG-4 movies)
```
...
<av.VideoFrame #110, pts=11000 yuv420p 4096x2160 at 0x15f9202b1c0>
<av.VideoFrame #111, pts=11100 yuv420p 4096x2160 at 0x15f9202ada0>
<av.VideoFrame #112, pts=11200 yuv420p 4096x2160 at 0x15f9202b9a0>
<av.VideoFrame #119, pts=11300 yuv420p 4096x2160 at 0x15f9202a620>
<av.VideoFrame #119, pts=11400 yuv420p 4096x2160 at 0x15f9202abc0>
<av.VideoFrame #119, pts=11500 yuv420p 4096x2160 at 0x15f9202ba00>
<av.VideoFrame #119, pts=11600 yuv420p 4096x2160 at 0x15f9202bb80>
<av.VideoFrame #119, pts=11700 yuv420p 4096x2160 at 0x15f9202bac0>
<av.VideoFrame #119, pts=11800 yuv420p 4096x2160 at 0x15f9202bb20>
<av.VideoFrame #119, pts=11900 yuv420p 4096x2160 at 0x15f9202b220>
```
Tail end frames get repeated numbers. timestamps are correct.
## Investigation
I tested on a simple video with frames 1-120 encoded from a wide array of source encoders, ffmpeg, pyav, handbrake.
```
import av
path = SOURCE

container = av.open(path, mode='r', format='mp4')
vid_stream = container.streams.video[0]
vid_stream.thread_type = 'AUTO'
frames = container.decode(vid_stream)

for frame in frames:
	print(frame)
```

## Reproduction

Load any video and print each frame. The results will illustrate the problem.

## Versions

- OS: Windows 10 Pro 10.0.19045 Build 19045
- PyAV runtime:  10.0.0
- PyAV build: ?
- FFmpeg:
```
[...]

Issue #33, opened 2014-07-09
frame index of video frame


mkassner:
Since frame.pts is often only by coincidence equal to frame index in video stream (and will ultimately fail with non constant frame rates), I was hoping for frame.index to represent the frame index in the steam.

However, it appears that frame index == encoded frame count on frame creation. This mean that after a seek the index become -at least the way I understand it- false.

Any ideas on how to fix that?

funkybob:
In my own code, after much painful and tedious research :) I have the following:

```
'''
        PTS (Presentation Time Stamp), which indicates when this frame should
        be shown, is expressed according to stream.time_base.

        We can compare the PTS to frame numbers if we scale by:
            (stream.average_rate / stream.time_base)
'''
pts_scale = video_stream.average_rate * video_stream.time_base

frame_pts = int(frame_index / pts_scale)

```

I'm not certain how portable this is, as we only use MP4 video streams.

jlaine:
I'm going to close this issue, as the following time-related attributes are properly documented nowadays:

- dts
- pts
- time_base

PR #1265, opened 2024-01-21
Prevent repeating frame indices

JoeSchiff:
Fix for Issue #1158.

#### Problem
`stream.thread_type.FRAME` threading causes an incorrect series of repeating frame indices at the end of most videos. This is because a list of frames is returned by [`CodecContext._send_packet_and_recv`](https://github.com/PyAV-Org/PyAV/blob/1962443425d9f63c80d220546ac5f3dfb0799f6f/av/codec/context.pyx#L405) . `frame.index` corresponds to the most recent frame received from the decoder. Therefore all frames in the list are assigned the same index as the final frame in the list.

#### Solution
I solved this issue by detecting whenever multiple frames are returned and then correcting the `frame.index` based on the frame's position in the list.

#### Other changes
`AVCodecContext::frame_number` [is deprecated](https://ffmpeg.org//doxygen/trunk/structAVCodecContext.html#a1606ae31b14af5a2940b94c99264c0fc). I switched to `AVCodecContext::frame_num`.
EDIT: The deprecation was issued in ffmpeg 6.0, so I reverted back to using `AVCodecContext::frame_number`.

#### Testing
The test iterations may be overkill, but it only increased the test run time from 6 seconds to 6.5 seconds on my machine.

#### Future considerations
It is important to note that `frame.index` is affected by `CodecContext.skip_frame`. Skipped frames will not increment the frame index. I don't know if this is intended or desired behavior. Perhaps we should consider two different frame properties.
1: A property for the number of frames returned by the decoder (this is the current behavior). 
2: A property for the frame index relative to the total number of frames in the video. I assume this can be determined by a timestamp?

### DIFF
--- a/av/codec/context.pxd
+++ b/av/codec/context.pxd
@@ -43,6 +43,7 @@ cdef class CodecContext:
     cdef _prepare_and_time_rebase_frames_for_encode(self, Frame frame)
     cdef _prepare_frames_for_encode(self, Frame frame)
     cdef _setup_encoded_packet(self, Packet)
+    cdef _correct_frame_indices(self, frame_list)
     cdef _setup_decoded_frame(self, Frame, Packet)
 
     # Implemented by base for the generic send/recv API.
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -513,23 +513,33 @@ cdef class CodecContext:
 
         self.open(strict=False)
 
-        res = []
+        frame_list = []
         for frame in self._send_packet_and_recv(packet):
             if isinstance(frame, Frame):
                 self._setup_decoded_frame(frame, packet)
-            res.append(frame)
-        return res
+            frame_list.append(frame)
+        if packet and packet.stream and packet.stream.type != "subtitle":
+            self._correct_frame_indices(frame_list)
+        return frame_list
+
+    cdef _correct_frame_indices(self, frame_list):
+        """
+        All frames in the list returned by '_send_packet_and_recv' will have the same index.
+        This will set the correct indices.
+        """
+        cdef Frame frame
+        correction = len(frame_list)
+        for frame in frame_list:
+            frame.index = self.ptr.frame_number - correction
+            correction -= 1
 
     cdef _setup_decoded_frame(self, Frame frame, Packet packet):
-
         # Propagate our manual times.
         # While decoding, frame times are in stream time_base, which PyAV
         # is carrying around.
         # TODO: Somehow get this from the stream so we can not pass the
         # packet here (because flushing packets are bogus).
         frame._time_base = packet._time_base
-
-        frame.index = self.ptr.frame_number - 1
 
     property name:
         def __get__(self):
--- a/tests/test_decode.py
+++ b/tests/test_decode.py
@@ -124,3 +124,22 @@ class TestDecode(TestCase):
                     getattr(container, attr)
                 except AssertionError:
                     pass
+
+    def test_decode_frame_indices(self):
+        for thread_type in ("NONE", "AUTO"):
+            for thread_count in range(4):
+                for skip_type in ("NONE", "NONKEY"):
+                    frame_list = []
+                    compare_list = []
+                    frame_count = 0
+                    with av.open(fate_suite("h264/interlaced_crop.mp4")) as container:
+                        stream = container.streams.video[0]
+                        stream.thread_type = thread_type
+                        stream.thread_count = thread_count
+                        stream.codec_context.skip_frame = skip_type
+                        for packet in container.demux(stream):
+                            for frame in packet.decode():
+                                frame_list.append(frame.index)
+                                compare_list.append(frame_count)
+                                frame_count += 1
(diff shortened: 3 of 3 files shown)


WyattBlue:
I would like there to be performance benchmarks ran on `decode`. Eyeballing the code, it looks this costs more cpu time but the PR is still good if the performance difference isn't measurable.

I'm curious why you choose to extract the logic of `_correct_frame_indices` to it's own function instead of inlining it in `decode`. This again feels like a performance thing but I would like measurable results. Actually, since `_setup_decoded_frame` is only one line now, that could be inlined as well.

jlaine:
I'm very doubtful about this PR, where is the behaviour you describe documented?

JoeSchiff:
#### Performance Benchmarks
@WyattBlue I ran some rudimentary performance benchmarks using the following code:
```
import av
from av.datasets import fate as fate_suite
import time

PATHS = (fate_suite("h264/bbc2.sample.h264"), fate_suite("h264/interlaced_crop.mp4"), av.datasets.curated("pexels/time-lapse-video-of-night-sky-857195.mp4"))

def test_decode_frame_indices():
    for path in PATHS:
        for thread_type in ("NONE", "AUTO"):
            for thread_count in range(4):
                for skip_type in ("NONE", "NONKEY"):
                    frame_list = []
                    compare_list = []
                    frame_count = 0
                    #print('\n', path, thread_type, thread_count, skip_type)
                    with av.open(path) as container:
                        stream = container.streams.video[0]
                        stream.thread_type = thread_type
                        stream.thread_count = thread_count
                        stream.codec_context.skip_frame = skip_type
                        for packet in container.demux(stream):
                            for frame in packet.decode():
                                frame_list.append(frame.index)
                                compare_list.append(frame_count)
                                frame_count += 1
                    print(frame_list == compare_list)

start_time = time.perf_counter()
test_decode_frame_indices()
print(time.perf_counter() - start_time)
```

I did 3 runs on each branch and got the following average durations:
```
main: 5.705269041995052 seconds
fix_frame_indices: 5.665727951010922 seconds
```
If you were referring to more sophisticated benchmarking, I'll need some help to get started.


#### Function vs Inline
The only reason I made a function was to make it slightly more organized. If you prefer that I change to inline, that's perfectly fine with me.

JoeSchiff:
#### AVFrame
@jlaine I looked through [AVFrame Struct Reference](https://ffmpeg.org//doxygen/trunk/structAVFrame.html) and found `display_picture_number`, which looked promising. However, when I exposed it in pyav it always returned 0. 
It is also deprecated in ffmpeg 6.0.

[This issue](https://github.com/PyAV-Org/PyAV/issues/33#issuecomment-48817817) seems to also conclude that `display_picture_number` is not trustworthy.

#### Docs
`frame.index` is found in the pyav.org docs [here](https://pyav.org/docs/stable/index.html?highlight=index#basic-demo) and [here](https://pyav.org/docs/stable/api/frame.html?highlight=index#av.frame.Frame.index).

jlaine:
If the point is purely to have a running index, what is gained over doing:

```python

for (index, frame) in enumerate(container.decode(video=0)):
   ...
```

JoeSchiff:
@jlaine Yeah, that's a good point. However, since `frame.index` is on the pyav.org homepage, I'm sure many people expect it to be available. I would really like to provide a replacement before removing it.

It seems these problems arise from the fact that `frame.index` is not an actual index. `enumerate()` and `frame.index` work correctly only when iterating every frame. They will break if using `CodecContext.skip_frame` or `container.seek`.

How about this solution:
1. The current functionality of "number of frames decoded so far" is moved from `frame.index` to a new property called `CodecContext.frame_num`. That way it matches ffmpeg and the original functionality is still available to users.

2. Create `frame.guessed_index` which is calculated from `frame.pts * stream.guessed_rate * frame.time_base` or similar (I still struggle with the intricacies of time base).

The problem is `frame.pts`, `stream.guessed_rate`, and `frame.time_base` are not always available in all media. Example: fate_suite("h264/bbc2.sample.h264")

If `frame.guessed_index`  isn't available then the original intent of this PR can be implemented by the user like this:
```
for packet in container.demux(stream):
    frame_list = packet.decode()
    correction = len(frame_list)
    for frame in frame_list:
        corrected_index = stream.codec_context.frame_num - correction
        correction -= 1
        print(f"{corrected_index=}")
```

Ideally we would be able to get a frame's index from ffmpeg, but I can't find a way to do it at the moment. I think this new solution solves many problems:
1. The original functionality of `frame.index` remains (elsewhere).
2. The repeating frame indices from [#1158](https://github.com/PyAV-Org/PyAV/issues/1158) can be solved.
3. Pyav matches ffmpeg's structure and property names.
4. Provide a way to get the frame index even when `CodecContext.skip_frame = "NONKEY"`.
5. The name `frame.guessed_index` makes it clear that the value is not guaranteed to be accurate.

WyattBlue:
I agree with jlaine. We should deprecate and remove `frame.index`. There is no such concept in ffmpeg. Putting in a `guessed_index` would only further complicate an already complex project.

PR #1266, opened 2024-01-21
Stop using `wheel` as an unconditional build dep

webknjaz:
It used to be an incorrect recommendation in some of the upstream docs. `setuptools` auto-injects it when building wheels and it's unnecessary when building sdist.


WyattBlue:
Thanks for the catch

PR #1279, opened 2024-02-16
Add support for ffmpeg's AVCodecContext::delay

JoeSchiff:
This :type: int `delay` variable is currently inaccessible:
https://github.com/PyAV-Org/PyAV/blob/8e32fe7ed7b83eff27682b4e026e7320cb8dfe1e/include/libavcodec/avcodec.pxd#L192

I assume this is supposed to be a reference to ffmpeg's [AVCodecContext::delay](https://ffmpeg.org/doxygen/6.0/structAVCodecContext.html#a948993adfdfcd64b81dad1151fe50f33).

This PR creates a property `codec_context.delay` which references ffmpeg's `AVCodecContext::delay`.

There is a separate, but similarly named, PyAV property `codec.delay` which is a reference to ffmpeg's [AV_CODEC_CAP_DELAY](https://ffmpeg.org/doxygen/6.0/group__lavc__core.html#ga3f55f5bcfbb12e06c7cb1195028855e6).
We might want to rename one of these properties to prevent confusion.

#### Tests
Since the ffmpeg docs for `AVCodecContext::delay` state "Set by libavcodec.", I was unsure how to test this property. I ended up finding a file in the FATE suite with a detected delay value. I verified the value (312 for this example) using ffprobe.

If this is an acceptable testing strategy, then I could write up more tests to help merge some of the other PRs.

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -640,3 +640,13 @@ cdef class CodecContext:
 
         def __set__(self, value):
             self.ptr.skip_frame = SkipType[value].value
+
+    property delay:
+        """Codec delay.
+
+        Wraps :ffmpeg:`AVCodecContext.delay`.
+
+        """
+
+        def __get__(self):
+            return self.ptr.delay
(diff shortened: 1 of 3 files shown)


WyattBlue:
Thanks! Looks go to me

PR #1287, opened 2024-02-22
Convert Cython extension class properties to decorator syntax part 1 of 4

JoeSchiff:
"Showing 24 changed files with 907 additions and 843 deletions."
I doubt that you want this all in one PR, so I broke it up into more manageable chunks.

We currently use a legacy syntax for defining properties in an extension class which has been deprecated since ~2016:
https://cython.readthedocs.io/en/stable/src/userguide/extension_types.html#properties

Switching to the new syntax required moving the property docstrings. Previously, the docstring was placed on the line below the `property` keyword. However, using the new syntax there can not be anything between the decorator and the `def` statement, so the docstring was moved to below the def statement.

### DIFF
--- a/av/descriptor.pyx
+++ b/av/descriptor.pyx
@@ -18,42 +18,42 @@ cdef class Descriptor:
         if sentinel is not _cinit_sentinel:
             raise RuntimeError("Cannot construct av.Descriptor")
 
-    property name:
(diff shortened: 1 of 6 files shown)


WyattBlue:
Thank you

PR #1289, opened 2024-02-24
Convert deprecated cython extension class properties to new syntax part 3

JoeSchiff:
WyattBlue, I saw a note you made here:
https://github.com/PyAV-Org/PyAV/wiki/TODO-general
"Convert properties to decorator syntax. This will also allow doctest to run on them."

Could you explain what you meant by this? Is doctest incompatible with the old syntax?

### DIFF
--- a/av/data/stream.pyx
+++ b/av/data/stream.pyx
@@ -15,9 +15,9 @@ cdef class DataStream(Stream):
     def decode(self, packet=None, count=0):
         return []
 
-    property name:
-        def __get__(self):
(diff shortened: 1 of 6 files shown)


WyattBlue:
> WyattBlue, I saw a note you made here:
> https://github.com/PyAV-Org/PyAV/wiki/TODO-general

Mike Boers wrote that, actually. I don't think those comments are relevant anymore

mikeboers:
I don’t read everything that goes on here anymore (duh), but when the notification looks like that I can’t refuse, lol

PR #1291, opened 2024-02-25
Convert deprecated cython extension class properties to new syntax part 4

JoeSchiff:
I had some trouble with the `color_range` property in av/video/codeccontext.pyx and tests/test_colorspace.py.

After my changes, `stream.color_range` now returns `2` instead of `None` and therefore fails the test. I have no idea which one is actually correct.

Here is some additional context which may or may not be related:
1. I get this warning when opening the file used in tests/test_colorspace.py:
`container = av.open(fate_suite("amv/MTV_high_res_320x240_sample_Penguin_Joke_MTV_from_WMV.amv"))`
`WARNING  libav.avi:test_colorspace.py:9 scale/rate is 0/0 which is invalid. (This file has been generated by broken software.)`

2. The `color_range` property in av/video/codeccontext.pyx had unusal syntax. The decorator was used, but so was `__get__` and `__set__`.

3. There are two `color_range` data fields in ffmpeg:
AVCodecContext::color_range
https://ffmpeg.org//doxygen/6.0/structAVCodecContext.html#a255bf7100a4ba6dcb6ee5d87740a4f35
AVFrame::color_range
https://ffmpeg.org//doxygen/6.0/structAVFrame.html#a853afbad220bbc58549b4860732a3aa5

I'll do more research in the next few days, unless the solution is obvious to someone already.

### DIFF
--- a/av/codec/codec.pyx
+++ b/av/codec/codec.pyx
@@ -209,17 +209,17 @@ cdef class Codec:
         from .context import CodecContext
         return CodecContext.create(self)
 
-    property is_decoder:
-        def __get__(self):
-            return not self.is_encoder
+    @property
+    def is_decoder(self):
+        return not self.is_encoder
 
-    property descriptor:
-        def __get__(self): return wrap_avclass(self.ptr.priv_class)
+    @property
+    def descriptor(self): return wrap_avclass(self.ptr.priv_class)
 
-    property name:
-        def __get__(self): return self.ptr.name or ""
-    property long_name:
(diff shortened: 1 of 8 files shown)


WyattBlue:
I think `2` is correct and `None` is wrong. The old color_range getter looks like it incorrectly always returns None.

JoeSchiff:
WyattBlue, testing with ffprobe confirms that you are correct. It feels good to unexpectedly squash that bug.

PR #1306, opened 2024-03-03
Add stub for buffer, descriptor, plane

laggykiller:


WyattBlue:
`    def __init__(self, sentinel: object) -> None: ...` is wrong. The user can't initialize that object so `__init__` shouldn't be in that type stub.

Also, you should always end files with a newline. It's part of our codestyle. GitHub displays a big red sign when a file doesn't.

laggykiller:
> The user can't initialize that object so __init__ shouldn't be in that type stub.

It is true that the user cannot initialize the object, I agree that I should make change on typing of `__init__`

However, I think we still need to make stub for those classes, as they could be accessed from other objects such as `av.audio.plane`.

laggykiller:
What if I change to this: https://github.com/laggykiller/PyAV/commit/88e9f0f82cb8757312570f60005a72bb83133a76

PR #1308, opened 2024-03-03
add missing stub in av.sidedata

laggykiller:
Requires: #1306

### DIFF
--- a/av/sidedata/motionvectors.pyi
+++ b/av/sidedata/motionvectors.pyi
@@ -1,0 +1,25 @@
+from typing import Any, Sequence
+
+import numpy as np
+
+from .sidedata import SideData
+
+class _MotionVector(SideData):
+    def __getitem__(self, index: int) -> MotionVector: ...
+    def __len__(self) -> int: ...
+    def to_ndarray(self) -> np.ndarray[Any, Any]: ...
+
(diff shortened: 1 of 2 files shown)


av/sidedata/motionvectors.pyi
```diff
@@ -0,0 +1,25 @@
+from typing import Any, Sequence
+
+import numpy as np
+
+from .sidedata import SideData
+
+class _MotionVector(SideData):
```
WyattBlue, line 7: _MotionVector is not supposed to be part of the Public API, so it's attributes should be merged with MotionVector

av/sidedata/sidedata.pyi
```diff
@@ -0,0 +1,33 @@
...
+    SPHERICAL: int
+    CONTENT_LIGHT_LEVEL: int
+    ICC_PROFILE: int
+    SEI_UNREGISTERED: int
+
+class SideData(Buffer):
+    type: Type
+    DISPLAYMATRIX: int
+
+class _SideDataContainer:
```
WyattBlue, line 30: Same applies here

laggykiller:
Please check, I am not sure if this fix is correct...

WyattBlue:
I think #1311 needs to be merged before this can.

PR #1311, opened 2024-03-03
Improve stub for enum, error, logging, stream

laggykiller:


WyattBlue:
I don't know if the stubs for enum, error, logging are even accurate. It's been a while since I reviewed that code and I know logging needs an overhaul in general.

stream.pyx shouldn't be included with these files.

laggykiller:
> I don't know if the stubs for enum, error, logging are even accurate.

I made the stub by reading the pyx files, as well as running pyav, open a video and check if those attributes could be accessed.

> stream.pyx shouldn't be included with these files.

Please reconsider about this. Even if it should not be accessed by the user, it is easy for us to create a stub file next to the pyx that is responsible for it. `Stream` from `stream.pyx` is baseclass for `VideoStream` and `AudioStream`. If we do not create stub for `stream.pyx`, further changes in `stream.pyx` (e.g. add/remove attributes from `Stream`) would require us to edit av.video.stream and av.audio.stream stub files.

av/enum.pyi
```diff
@@ -18,17 +29,17 @@ class EnumItem:
...
     flags: tuple[EnumFlag]
 
-    def __and__(self, other): ...
-    def __or__(self, other): ...
-    def __xor__(self, other): ...
-    def __invert__(self): ...
+    def __and__(self, other: object) -> None: ...
+    def __or__(self, other: object) -> None: ...
+    def __xor__(self, other: object) -> None: ...
+    def __invert__(self) -> None: ...
```
WyattBlue, line 42: This looks sus to me. Are you sure it's `None` and not `bool`?

av/stream.pyi
```diff
@@ -22,5 +32,5 @@ class Stream:
     language: str | None
     type: Literal["video", "audio", "data", "subtitle", "attachment"]
 
-    def encode(self, frame=None): ...
-    def decode(self, packet=None): ...
+    def encode(self, frame: Frame | None = None) -> list[Packet]: ...
+    def decode(self, packet: Packet | None = None) -> list[Frame]: ...
```
WyattBlue, line 36: Make sure these methods overloaded in VideoStream and AudioStream respectively

av/enum.pyi
```diff
@@ -18,17 +29,17 @@ class EnumItem:
 
     def __int__(self) -> int: ...
     def __hash__(self) -> int: ...
-    def __reduce__(self): ...
-    def __eq__(self, other) -> bool: ...
-    def __ne__(self, other) -> bool: ...
+    def __reduce__(self) -> tuple[Any, ...]: ...
```
WyattBlue, line 32: `tuple[bool, ...]` ???

av/enum.pyi
laggykiller: You are right. I bulk edited the return type without thinking...

av/stream.pyi
laggykiller: Yes they are, no worries

av/enum.pyi
laggykiller: Thanks!

laggykiller:
Fixed, please review!

av/enum.pyi
```diff
@@ -18,17 +29,17 @@ class EnumItem:
...
     flags: tuple[EnumFlag]
 
-    def __and__(self, other): ...
-    def __or__(self, other): ...
-    def __xor__(self, other): ...
-    def __invert__(self): ...
+    def __and__(self, other: object) -> bool: ...
+    def __or__(self, other: object) -> bool: ...
+    def __xor__(self, other: object) -> bool: ...
+    def __invert__(self) -> bool: ...
```
WyattBlue, line 42: Actually, the return type should be EnumFlag, not bool. Looking at test_enums.py

av/enum.pyi
```diff
@@ -18,17 +29,17 @@ class EnumItem:
 
     def __int__(self) -> int: ...
     def __hash__(self) -> int: ...
-    def __reduce__(self): ...
-    def __eq__(self, other) -> bool: ...
-    def __ne__(self, other) -> bool: ...
+    def __reduce__(self) -> tuple[bool, ...]: ...
```
WyattBlue, line 32: I actually have no clue what `__reduce__` returns, or if this method is even needed at all. Just use set it to `Any`

av/enum.pyi
laggykiller: I think you are right

av/enum.pyi
laggykiller: Should be `tuple[Callable[[str, str, str], Any], tuple[str, str, str]]`.

`enum.pyx`:
```python
    def __reduce__(self):
        return (_unpickle, (self.__class__.__module__, self.__class__.__name__, self.name))
```

```python
def _unpickle(mod_name, cls_name, item_name):
    mod = __import__(mod_name, fromlist=["."])
    cls = getattr(mod, cls_name)
    return cls[item_name]
```

laggykiller: Mystery solved:

```python
>>> from av.video.frame import PictureType
>>> t = PictureType
>>> t
<class 'av.video.frame.PictureType'>
>>> t[0]
<av.video.frame.PictureType:NONE(0x0)>
>>> t[0].__reduce__()
(<cyfunction _unpickle at 0x73e2344805f0>, ('av.video.frame', 'PictureType', 'NONE'))
>>> t[0].__reduce__()[0](*t[0].__reduce__()[1])
<av.video.frame.PictureType:NONE(0x0)>
>>> type(t[0].__reduce__()[0](*t[0].__reduce__()[1]))
<class 'av.video.frame.PictureType'>
```

So it should be tuple[Callable[[str, str, str], EnumItem], tuple[str, str, str]]

PR #1316, opened 2024-03-03
Improve stub in av.audio

laggykiller:
Requires: #1306

### DIFF
--- a/av/audio/stream.pyi
+++ b/av/audio/stream.pyi
@@ -1,3 +1,12 @@
+from av.packet import Packet
 from av.stream import Stream
 
-class AudioStream(Stream): ...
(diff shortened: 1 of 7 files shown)


av/audio/stream.pyi
```diff
@@ -1,3 +1,12 @@
+from av.packet import Packet
 from av.stream import Stream
 
-class AudioStream(Stream): ...
+from .format import AudioFormat
+from .frame import AudioFrame
+
+class AudioStream(Stream):
+    format: AudioFormat
+    type = "audio"
```
WyattBlue, line 9: Should be type: Literal["audio"]
Looks good to me otherwise.

PR #1318, opened 2024-03-03
Improve stubs in av.container

laggykiller:


av/container/output.pyi
```diff
@@ -1,5 +1,21 @@
...
+
 from .core import Container
 
 class OutputContainer(Container):
     def __enter__(self) -> OutputContainer: ...
+    def add_stream(
+        self,
+        codec_name: str,
+        rate: Fraction | int | float,
+        template: Stream,
```
WyattBlue, line 15: Should be `template: Stream | None = None`

laggykiller: I see that either codec_name or template is needed. Should I type it as an overload function instead?

WyattBlue: Yes, overload is appropriate here

laggykiller: Sorry I edited the previous comment, I found out that typing as overloading might be impossible

PR #1327, opened 2024-03-08
Getter tests for packet properties

JoeSchiff:
Tests to help finish #1138.

I searched through the FATE suite and I was able to find files with `corrupt`, `discard`, and `disposable` packet flags. I was unable to find a file with a `trusted` flag.
I then used ffprobe to determine the packet indices for `keyframe`, `corrupt`, and `discard` flags in their respective files, but I could not detect the `disposable` flag. 
I'm guessing ffprobe doesn't provide that flag by default. Does anyone know how to get more details about flags?
These are the commands I tried:
`ffprobe -show_entries packet=flags "$file"`
`ffprobe -show_packets "$file"`


WyattBlue:
Is homebrew breaking our tests?

WyattBlue:
I think FATE is down. The tests pass on my local machine so it looks good.

PR #1358, opened 2024-03-25
Remove encode and decode for DataStream

WyattBlue:


dotysan:
Why?

WyattBlue:
Because the implementation only returns blank data.

PR #565, opened 2019-10-26
Bitstream filters

mikeboers:

Issue #489, opened 2019-02-26
Using Bit Stream Filters


hunterjm:
I am trying to convert an MP4 H264 stream to Annex-B for use in a MPEGTS output container.  I am using the MP4 generated video file located in the docs under `numpy` for my reproduction example.  I am getting an error saying that the stream is invalid, and needs to be in Annex B using `-bsf:v h264_mp4toannexb`

### Error
```
H.264 bitstream malformed, no startcode found, use the video bitstream filter 'h264_mp4toannexb' to fix it ('-bsf:v h264_mp4toannexb' option with ffmpeg)
```

### Reproduction code:
```python
import io

import numpy as np
import av

duration = 4
fps = 24
total_frames = duration * fps

mp4 = io.BytesIO()
mp4.name = 'test.mp4'
container = av.open(mp4, mode='w')

stream = container.add_stream('libx264', rate=fps)
stream.width = 480
stream.height = 320
stream.pix_fmt = 'yuv420p'

for frame_i in range(total_frames):

    img = np.empty((480, 320, 3))
    img[:, :, 0] = 0.5 + 0.5 * np.sin(2 * np.pi * (0 / 3 + frame_i / total_frames))
    img[:, :, 1] = 0.5 + 0.5 * np.sin(2 * np.pi * (1 / 3 + frame_i / total_frames))
    img[:, :, 2] = 0.5 + 0.5 * np.sin(2 * np.pi * (2 / 3 + frame_i / total_frames))

    img = np.round(255 * img).astype(np.uint8)
    img = np.clip(img, 0, 255)

    frame = av.VideoFrame.from_ndarray(img, format='rgb24')
    for packet in stream.encode(frame):
        container.mux(packet)

# Flush stream
for packet in stream.encode():
    container.mux(packet)

# Close the file
container.close()


# Convert to mpegts
input_ = av.open(mp4, 'r')
input_stream = input_.streams.video[0]
output_file = io.BytesIO()
output_file.name = 'test.ts'
output = av.open(output_file, 'w', format='mpegts')
output_stream = output.add_stream('h264', input_stream.rate)

for packet in input_.demux(input_stream):
    # We need to skip the "flushing" packets that `demux` generates.
    if packet.dts is None:
        continue

    # We need to assign the packet to the new stream.
    packet.stream = output_stream

    output.mux(packet)

output.close()
```

jlaine:
Hi Jason!

[...]

PR #1375, opened 2024-04-08
Add bitstream filters

skeskinen:
Loosely based on #565
Additional discussion in #489 

This adds bindings for AVBSFContext (BitStreamFilterContext)

I decided against adding binding for AVBitStreamFilter, because it seems like a little bit useless API. You can't do anything with such object besides creating a new BSF context. (except checking which codecs are supported for that BSF, I guess)
It makes the code simpler and imo easier to maintain. Let me know if this is a mistake and those bindings should be added as well.

I think it makes sense to force BitstreamFilterContext creation through the string parsing API, since that's the only way I can see how to set filter options and also allows things like filter chains, etc. 

Adds tests for 'chomp', 'setts' and 'h264_mp4toannexb' filters. We could also test that filter chains work.

h264_mp4toannexb requires that AVCodecParameters *par_in is set correctly. To that end BitStreamFilterContext takes a Stream as an argument and copies the codec parameters from there.

There are also attributes `time_base_in`, `time_base_out` which are probably required for 100% support of setts option constraints. Let me know if those should be hooked up as well. I guess it would require that BitStreamFilterContext takes separate in_stream, out_stream arguments instead of single stream.

### DIFF
--- a/av/__init__.py
+++ b/av/__init__.py
@@ -29,6 +29,7 @@ from av.audio.format import AudioFormat
 from av.audio.frame import AudioFrame
 from av.audio.layout import AudioLayout
 from av.audio.resampler import AudioResampler
+from av.bitstream import BitStreamFilterContext, bitstream_filters_available
 from av.codec.codec import Codec, codecs_available
 from av.codec.context import CodecContext
 from av.container import open
--- a/av/bitstream.pxd
+++ b/av/bitstream.pxd
@@ -2,6 +2,7 @@ cimport libav as lib
 
 from av.packet cimport Packet
 
+
 cdef class BitStreamFilterContext:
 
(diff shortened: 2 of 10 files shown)


skeskinen:
Excuse me about the linter stuff. I can't run make lint locally because the flake8 always fails. I think it should work now. Both black and isort are happy

WyattBlue:
I think the PR is good as it is now.
One thing I changed was inlining `def self._recv` and `def self._send` to `cpdef filter`. This eliminates the overhead in calling Python functions and IMO makes it easier to read. I also add type hints and docstrings.

I'll give you a chance to comment, then it can be merged.

skeskinen:
Looks good to me. Thanks!

PR #1383, opened 2024-04-24
Disable logs by default

WyattBlue:


WyattBlue:
We're affected by https://github.com/actions/setup-python/issues/852 which is why our Mac builds are falling.

Issue #1267, opened 2024-01-30
Allow way to get real aspect ratio


animetosho:
## Overview

`video.display_aspect_ratio` and `video.sample_aspect_ratio` don't always return the correct values.

## Reproduction

Using the following script:

```python
import av
import sys

container = av.open(sys.argv[1])
video = next(s for s in container.streams if s.type == 'video')
print(video.display_aspect_ratio)
print(video.sample_aspect_ratio)
```

[Sample file 1](https://storage.animetosho.org/sframes/000fd474_9930.mkv) output:

	# python3 test.py 000fd474_9930.mkv
	4/3
	3/4
	# ffprobe -hide_banner 000fd474_9930.mkv
	Input #0, matroska,webm, from '000fd474_9930.mkv':
	  Metadata:
	    ENCODER         : Lavf58.76.100
	  Duration: 00:00:00.04, bitrate: 47051 kb/s
	  Stream #0:0: Video: h264 (High), yuv420p(progressive), 1920x1080 [SAR 3:4 DAR 4:3], SAR 1:1 DAR 16:9, 23.98 fps, 23.98 tbr, 1k tbn, 47.95 tbc (default)
	    Metadata:
	      DURATION        : 00:00:00.041000000

[Sample file 2](https://storage.animetosho.org/sframes/00115c29_11730.mkv) output:

	# python3 test.py 00115c29_11730.mkv
	None
	None
	# ffprobe -hide_banner 00115c29_11730.mkv
	[libdav1d @ 0x561c7adb3280] libdav1d 0.9.1
	Input #0, matroska,webm, from '00115c29_11730.mkv':
	  Metadata:
	    ENCODER         : Lavf58.76.100
	  Duration: 00:00:00.04, start: 0.000000, bitrate: 45895 kb/s
	  Stream #0:0: Video: av1 (Main), yuv420p10le(tv, bt709), 1440x1080 [SAR 1:1 DAR 4:3], SAR 4:3 DAR 16:9, 23.98 fps, 23.98 tbr, 1k tbn, 1k tbc (default)
	    Metadata:
	      DURATION        : 00:00:00.041000000
	[libdav1d @ 0x561c7adb62c0] libdav1d 0.9.1

It looks like these files have *two* aspect ratios (stream and container ratios perhaps?), with the second being the 'correct' one, but PyAV reports the first.  
[...]

PR #1425, opened 2024-06-17
fix: use `av_guess_sample_aspect_ratio` when reporting stream SAR

moonsikpark:
When user requests stream's SAR, use the upstream's heuristic function ([`av_guess_sample_aspect_ratio()`](https://github.com/FFmpeg/FFmpeg/commit/9163faecd3cdd93b69ae98605e7f518bd228196e)) to report it.

Resolves: https://github.com/PyAV-Org/PyAV/issues/1267


jlaine:
I'd really prefer we don't merge any new code without the corresponding tests.

PR #1432, opened 2024-06-21
Remove incorrect `av/__init__.pyi`

jlaine:
We do not need this type stub, since `__init__.py` is already a Python file. What's more the file is incorrect since it uses star imports, which the actual `__init__.py` does not.


WyattBlue:
I'm glad that `__init__.pyi` can be removed. However, that does mean we need to write a stub for `audio/fifo` since mypy is checking `__init__.py` now. I actually went and already did that and applied your changes #1433, so we don't need this pull request now.

PR #1442, opened 2024-06-25
Add Graph.vpush() and Graph.vpull()

WyattBlue:


jlaine:
No tests, no docstrings, no changelog entry?

What does this do and why do we need it?

WyattBlue:
Mostly to solve `(expression has type "VideoFrame | AudioFrame", variable has type "VideoFrame")` type errors when using the normal push and pull without having to use cast.

Issue #1378, opened 2024-04-16
AV 12.0.0 decoded frames do not match source as in AV 11.0.0 (potential colorspace issue)


mark-oshea:
## Overview

Decoding image frames from media which does not have a colorspace tag does not produce images that visually matche the source movie file.

Historically using both ffmpeg and pyav I have needed to provide the following color conversion when extracting YUV video frames to an RGB format. It's been this way for at least 5 years.

ffmpeg video filter: `-vf "colormatrix=bt709:bt601"`
pyav video reformatter args: `{"src_colorspace": "ITU709", "dst_colorspace": "ITU601"}`

This has been working fine for me in pyav 10.0.0 and 11.0.0 however as of 12.0.0 this no longer works and I cannot get my source video frames to match their RGB counterpart visually.

I believe the colormatrix/colorspace conversion is required due to libav defaulting to bt601 for video files which do not have a colorspace tag.

## Expected behavior

Decoded video frames should match the source video frames.

## Actual behavior

Decoded video frames do not match the source video frames.

## Investigation

Tested the same code with pyav 11.0.0 and pyav 12.0.0. Extracted frames visually match with av 11.0.0 but do not with av 12.0.0 when executing the same python script.

## Reproduction

Minimum example script below - note generally I want to extract video frames to numpy arrays for further processing external to pyav which is why I'm converting to an array here and saving using PIL. The roundtrip from `av.VideoFrame -> ndarray -> av.VideoFrame` should produce colors that match the source and historically they have.

I did try without the video_reformatter_args and without them, I do not get a visual match from either av 11 or av 12.

Media produced by script (along with source video file) can be accessed here:
https://drive.google.com/drive/folders/1OHHkJVP-6sHm9eQQ9iardT8VtwjVz0YT?usp=sharing

Loading up the files above in any media player which supports video and images will show that the image produced by av 11 matches the source video whilst the image produced by av 12 does not.

Example script:
```python
import av
from PIL import Image


SOURCE = "trainspotting.mov"
TARGET = f"trainspotting_frame_{av.__version__}.png"

container = av.open(SOURCE)
stream = container.streams.video[0]

# Needed for saved frame color to match source movie frame
video_reformatter_args = {"src_colorspace": "ITU709", "dst_colorspace": "ITU601"}

# Decode first frame and get 32bit array
frame = next(container.decode(stream))
pixels = frame.to_ndarray(format="gbrpf32le", **video_reformatter_args)

[...]

PR #1458, opened 2024-07-16
fix: set default color range for VideoReformatter.format()

elxy:
The behavior of the VideoReformatter.format() is inconsistent with the documentation. According to the documentation, when color_range is None, it should use the color_range of the frame. However, in the function implementation, the default is set to 0.


WyattBlue:
This reverts the fix to https://github.com/PyAV-Org/PyAV/issues/1378
Why don't we change the documentation instead?

elxy:
> This reverts the fix to #1378 Why don't we change the documentation instead?

Thanks for the reminder. The description of parameters has been updated to match the implementation.

Issue #1454, opened 2024-07-14
Error calling subtitle_stream.decode()


WyattBlue:
Minimum repro:
```python
# Any input that has a subtitle stream, codec does not matter
src = "/Users/wyattblue/projects/auto-editor/resources/subtitle.mp4"

c = av.open(src)
s = c.streams.subtitles[0]

for subset in s.decode():
    pass
```

```
zsh: segmentation fault  python3 test.py
```

Calling decode like this sidesteps the segfault and works as expected. This confirms this is our fault and not a downstream lib.
```python
for subset in c.decode(subtitles=0):
    for sub in subset:
        print(sub)
```

## Versions
OS: All (confirmed macos-arm64 macos-x86_64)
Python: All (confirmed 3.12, 3.11)
PyAVs: from at least 10.0.0, til 12.3.0rc1

moonsikpark:
```
(lldb) run test.py
Process 21360 launched: '/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/Resources/Python.app/Contents/MacOS/Python' (arm64)
Process 21360 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x0000000102804c2c libavcodec.60.31.102.dylib`avcodec_decode_subtitle2(avctx=0x000000014862f800, sub=0x0000000104282710, got_sub_ptr=0x000000016fdfdf54, avpkt=0x0000000000000000) at decode.c:972:9
   969  int avcodec_decode_subtitle2(AVCodecContext *avctx, AVSubtitle *sub,
   970                               int *got_sub_ptr, const AVPacket *avpkt)
   971  {
-> 972      int ret = 0;
   973 
   974      if (!avpkt->data && avpkt->size) {
   975          av_log(avctx, AV_LOG_ERROR, "invalid packet: NULL data, size != 0\n");
Target 0: (Python) stopped.
(lldb) frame variable avpkt
(const AVPacket *) avpkt = NULL
(lldb) kill
Process 21360 exited with status = 9 (0x00000009) killed
(lldb) run test2.py
Process 21617 launched: '/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/Resources/Python.app/Contents/MacOS/Python' (arm64)
Process 21617 stopped
[...]

PR #1461, opened 2024-07-18
flushing subtitle decoder requires a new uninitialized packet

moonsikpark:
Implement proper flushing mechanism when using `avcodec_decode_subtitle2`.

Closes: #1454


moonsikpark:
It would be nice if there is something in the fate suite which would allow us to test the flushing action.

WyattBlue:
Good job! I think you really went above and beyond

Issue #1482, opened 2024-08-07
Expose `av_buffersink_set_frame_size`


rectalogic:
## Overview

Expose `av_buffersink_set_frame_size` for use on audio filter graph `abuffersink`s

## Existing FFmpeg API
AudioResampler uses [`av_buffersink_set_frame_size`](https://ffmpeg.org/doxygen/trunk/group__lavfi__buffersink.html#ga359d7d1e42c27ca14c07559d4e9adba7) internally on it's internal filter graph `abuffersink`.
https://github.com/PyAV-Org/PyAV/blob/f5ec8ca837c41fb8905f2c1cd76e64c2c93c3624/av/audio/resampler.pyx#L98

I need a more complex graph that also uses `atrim` etc. but I still want to constrain the frame size. So it would be nice if `av_buffersink_set_frame_size` was exposed for applications to use directly.

## Expected PyAV API
Maybe there should be a new `add_abuffersink()` API on `Graph`, similar to `add_abuffer()`, that allows more configuration of the sink?

```python
graph = av.filter.Graph()
abuffersink = graph.add_abuffersink(frame_size=1024)
```

It might make sense to add `add_buffersink()` for video too - both of these sinks have non-str options and I'm not sure how they are configured using the pyav API currently.

```pycon
>>> graph.add("buffersink").filter.options
(<av.Option pix_fmts (BINARY at *0x10) at 0x108a70fd0>,)

>>> graph.add("abuffersink").filter.options
(<av.Option sample_fmts (BINARY at *0x20) at 0x108a70e20>, <av.Option sample_rates (BINARY at *0x60) at 0x108a71060>, <av.Option channel_layouts (BINARY at *0x30) at 0x108a71090>, <av.Option channel_counts (BINARY at *0x40) at 0x108a710c0>, <av.Option ch_layouts (STRING at *0x50) at 0x108a710f0>, <av.Option all_channel_counts (BOOL at *0x58) at 0x108a71120>)
```

Alternatively just expose the api directly, on Graph. `av_buffersink_set_frame_size` doesn't seem to care what kind of filter context it is called on, or even if it is an audio filter.
```python
graph = av.filter.Graph()
sink = graph.add("abuffersink")
graph.buffersink_set_frame_size(sink, 1024)
```

PR #1484, opened 2024-08-08
Add support for av_buffersink_set_frame_size

rectalogic:
Fixes https://github.com/PyAV-Org/PyAV/issues/1482


av/filter/graph.pyx
```diff
@@ -171,6 +171,15 @@ cdef class Graph:
 
         return self.add("abuffer", name=name, **kwargs)
 
+    def set_audio_frame_size(self, frame_size):
```
WyattBlue, line 174: Could use a docstring

PR #1519, opened 2024-09-05
Just try to build with ffmpeg 6

hmaarrfk:
I'm just trying to run the CIs with FFMPEG 6.

xref: https://github.com/PyAV-Org/PyAV/discussions/1518


hmaarrfk:
hmm, it seems your standard tests do run on FFMPEG 6.1.1...

PR #1533, opened 2024-09-19
feat: expose opaque and opaque_ref

CarlosRDomin:
Allows tagging AVPackets and AVFrames with any Python object and, when AV_CODEC_FLAG_COPY_OPAQUE is enabled, ffmpeg will take care of propagating the references to those objects from packets -> frames (decoders) and frames -> packets (encoders).

Example usage:
```python
import av
import time

container = av.open("video.mp4")
video_stream = container.streams.video[0]
video_stream.codec_context.copy_opaque = True  # Enable propagating `opaque` from packets to frames
for packet_idx, packet in enumerate(container.demux()):
    packet.opaque = (time.time(), packet_idx)  # Could be any Python object (dataclass, str, float, etc)
    frames = packet.decode()
    print(f"Packet #{packet_idx} yielded {len(frames)} frames")
    for frame in frames:
        print(f"Decoded frame with opaque={frame.opaque}")
```

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -69,6 +69,14 @@ Flags = define_enum("Flags", __name__, (
     ("DROPCHANGED", 1 << 5,
         "Don't output frames whose parameters differ from first decoded frame in stream."
     ),
+    ("RECON_FRAME", lib.AV_CODEC_FLAG_RECON_FRAME, "Request the encoder to output reconstructed frames, i.e. frames that would be produced by decoding the encoded bistream."),
+    ("COPY_OPAQUE", lib.AV_CODEC_FLAG_COPY_OPAQUE,
+        """Request the decoder to propagate each packet's AVPacket.opaque and AVPacket.opaque_ref
+        to its corresponding output AVFrame. Request the encoder to propagate each frame's
+        AVFrame.opaque and AVFrame.opaque_ref values to its corresponding output AVPacket."""),
+    ("FRAME_DURATION", lib.AV_CODEC_FLAG_FRAME_DURATION,
+        """Signal to the encoder that the values of AVFrame.duration are valid and should be
+        used (typically for transferring them to output packets)."""),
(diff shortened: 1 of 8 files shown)


WyattBlue:
What inspired you to make this PR? What's the use case for this?
This PR looks good, but we like to see tests, proper type stubs, and maybe some docs.

CarlosRDomin:
Hi @WyattBlue! Good point, I probably didn't provide much context, sorry about that.

I work with multi-modal data and it is super helpful to create real-time visualizations of e.g. an accelerometer signal overlaid on top of the video (or print any debugging text etc). In order to do that, I need to fuse the different sources and align them in time. The challenge is that video decoding has lag, and doesn't necessarily guarantee that the packets that come in, will map 1:1 to the frames that come out *in the same order*. My architecture looks something like:
 - A small service demuxing an rtsp stream, timestamping it, and sending the encoded h264/h265 packets through the network, along with some metadata (e.g. time and camera ID)
  - A service that parses the other sources of data that want to be visualized, and sends them in some common format
  - A service that receives the data from the 2 services above, time syncrhonizes it (e.g. this accelerometer packet goes with this video packet), decodes the video, passes it through a simple filterchain to render what we want, and visualizes it on the screen

With this PR, we can propagate the custom python dataclass containing the video timestamp, the camera ID, the accelerometer data, etc. from the video packets into the decoded frames without the need to implement some external deque or make assumptions about the order in which packets will come out, the `opaque` reference gets propagated natively by ffmpeg under the hood and everything just works :)


Regarding tests, I'm happy to add some, just wanted to circulate the idea first and see it would be an interesting contribution to PyAV.

Regarding type stubs, do you have an automated script to regenare the *.pyi files or should I edit them manually?

Thanks!

WyattBlue:
Thanks for the info. I'll merge once a test is written. For the type stubs, you'll need to edit them manually.

CarlosRDomin:
Awesome, will try to submit the tests on Monday :)

WyattBlue:
Merging now to prevent conflicts later.

PR #1554, opened 2024-09-25
If Conda needs this, they can patch it themselves

WyattBlue:


hmaarrfk:
The title of this PR is a little hostile.

I'll gladly provide input on behalf of conda packagers if you need some in the future.

hmaarrfk:
but I agree, we can patch things ourselves, but it does help to collaborate....

PR #1570, opened 2024-10-10
Stop using AVStream.side_data

WyattBlue:


hmaarrfk:
This seems premature right?

lgeiger:
> This seems premature right?

I'd agree. Although since only `DISPLAYMATRIX` was exposed and the [fact that `AVStream.side_data` seems to be undocumented](https://github.com/PyAV-Org/PyAV/issues/1150#issuecomment-2406037847) it does make sense to remove it in favour of consistency. It might make sense to expose [`AvPacket.side_data`](https://ffmpeg.org/doxygen/trunk/structAVPacket.html#ac55bfef91c33f02704ba76518d0f294c) in the future instead.

The problem for me was that this PR removed the only way to read the display rotation. However, #1675 brought back access to the rotation in a nicer API so this functionality will be accessible again in the next release.

hmaarrfk:
> brought back access to the rotation in a nicer API so this functionality will be accessible again in the next release.

Understood.

PR #1571, opened 2024-10-10
CFLAGS in Makefile does not exist

henri-gasc:
A quick search in GCC source code show that `-Wno-incompatible-function-pointer-types` does not exist as a flag. However, there is the close flag `-Wno-incompatible-pointer-types`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 LDFLAGS ?= ""
-CFLAGS ?= "-O0 -Wno-incompatible-pointer-types"
+CFLAGS ?= "-O0 -Wno-incompatible-pointer-types -Wno-unreachable-code"
 
 PYAV_PYTHON ?= python
 PYAV_PIP ?= pip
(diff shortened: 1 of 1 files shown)


henri-gasc:
Here is a line from my build log (after it failed to build version 13.1.0):
`cc1: note: unrecognized command-line option ‘-Wno-incompatible-function-pointer-types’ may have been intended to silence earlier diagnostics`

WyattBlue:
```
clang: error: unsupported option '--Wno-incompatible-pointer-type'
```
This PR isn't ready as is.

WyattBlue:
These flags are for both clang and GCC. Since clang does actually support `-Wno-incompatible-pointer-types`, I suppose this PR is fine. Although, can you add `-Wno-unreachable-code` if it works on GCC too?

henri-gasc:
Sorry, I did not think about clang when doing the change.  
I will test with `-Wno-unreachable-code`

henri-gasc:
There was no problem with it

PR #1610, opened 2024-11-02
Improve OutputContainer.add_stream overload

cdce8p:
Use TypeVar to improve `OutputContainer.add_stream` overload.


cdce8p:
Wondered for a moment why you closed the PR until I saw that it was merged as part of #1613. A short note would be nice next time 😃

This PR also removed an impossible overload. Since that wasn't part of #1613, I opened a separate PR for it:
- #1614

PR #1611, opened 2024-11-02
Add __all__ to audio and video packages

cdce8p:


WyattBlue:
I would prefer the '__all__' statements to be on one line.

cdce8p:
> I would prefer the '**all**' statements to be on one line.

Thought separate lines might be more extendible in the future but either option works. Pushed a new commit.

PR #1614, opened 2024-11-03
Remove impossible add_stream overload

cdce8p:
Remove overload which allowed `codec_name` and `template` to both have a value / have no value. The code requires that exactly one of the two has a value which is expressed by the other three overloads.

Leftover from #1610.

https://github.com/PyAV-Org/PyAV/blob/f8a07d0b83e2ec1479572542c168b79d7551d817/av/container/output.pyx#L46-L62


WyattBlue:
I think https://github.com/PyAV-Org/PyAV/pull/1615 solves this more completely.

cdce8p:
> I think #1615 solves this more completely.

Yes it does. Thanks!

PR #1618, opened 2024-11-06
Add qmin and qmax parameters to the codec context

davidplowman:
Allows applications to control the min and max quantisers used for encoding.

Also add a test to ensure the encoder is behaving as we expect.

(Just to explain the context here a bit, we're making much more use of PyAV in our standard Raspberry Pi software, so hopefully this will generate quite a few new users. I'm keen to get the tweaks and small new features on our fork back onto mainline here, because that's better for everyone, so there will probably be "a few" small pull requests like this. Thanks!)

### DIFF
--- a/av/video/codeccontext.pyi
+++ b/av/video/codeccontext.pyi
@@ -26,6 +26,8 @@ class VideoCodecContext(CodecContext):
     color_primaries: int
     color_trc: int
     colorspace: int
+    qmin: int
+    qmax: int
     type: Literal["video"]
 
     def encode(self, frame: VideoFrame | None = None) -> list[Packet]: ...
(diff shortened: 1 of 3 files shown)


WyattBlue:
You need to add type stubs. You might want to think about when the user tries to access it when it's a decoder.
Other than that, it looks good.

davidplowman:
> You need to add type stubs. You might want to think about when the user tries to access it when it's a decoder. Other than that, it looks good.

Hi, thanks for the quick reply! Just a couple of questions...

1. By "type stubs" do you mean putting new entries into the corresponding `codeccontext.pyi` file? I'll push an update with that change!

2. I was wondering about the decoder side of things. So far as I can tell, I don't think the decoder knows directly about `qmin` and `qmax` - they're just directives to the encoder, does that sound right? In principle I would have expected the decoder to be able to tell you the quantiser used for a frame (which would obviously bound the limits the encoder must have had), but I couldn't see it in the underlying `AVFrame` struct. Have I maybe missed something?

Thanks!

PR #1619, opened 2024-11-07
Bundle libsrt (support SRT protocol)

JoeSchiff:
This will solve: https://github.com/PyAV-Org/PyAV/discussions/1503

### Notes
SRT requires GnuTLS. Is there a cleaner way to install the package without globbing?

### Testing
I would like to set up a test for this, but I don't know how to start streaming a video from within pyav. I was able to create a network live stream with ffmpeg like this:
`ffmpeg -i local_file.mp4 -c:v libx264 -f mpegts 'srt://127.0.0.1:40052?mode=listener&latency=2000000'`

I then opened the network stream in pyav with:
```
path = "srt://127.0.0.1:40052?mode=caller"
container = av.open(path)
```
Pyav decoded the network stream and everything seemed to work.

If anyone knows how to create a network stream in pyav, I could write a test for this.


WyattBlue:
It looks like PyAV already supports SRT (at least for decoding) if the user compiles libSRT in. This code installs libsrt for testing, but you don't need that right unless you have a srt protocol test.

Look at https://github.com/PyAV-Org/PyAV-ffmpeg if you want to bundle libSRT in the pypi binary wheels.

JoeSchiff:
Thanks for the feedback. I'll check it out. 

Do you think an srt protocol test is necessary (or feasible)?

WyattBlue:
> Thanks for the feedback. I'll check it out. 
> 
> 
> 
> Do you think an srt protocol test is necessary (or feasible)?

Feasible? Yes.
Necessary? No, but a nice to have.

PR #1662, opened 2024-12-02
Ensure to add the loudnorm_impl.h in the pypi source distribution

hmaarrfk:
Thanks!


hmaarrfk:
Not sure if you can trigger a point release easily. I’m having more trouble patching this in than I thought tbh.

PR #1675, opened 2024-12-05
Add `VideoFrame.rotation`

### LINKED ISSUE #1047: Allow access to an affine transform of a particular video tracks display rotation.
## Overview

Hi. Firstly, apologies if this is a dupe. I did some cursory searching and found this thread which indicates a 'roll it yourself' POV. 

https://github.com/PyAV-Org/PyAV/issues/570

I'd like to propose an intermediary solution that puts the onus on the user for doing anything with the transform, but exposes it to clients of PyAV

Use case:

As a client of PyAV building a robust video processing system which consumes arbitrary videos, its a requirement to handle track transforms as we decode video within pyAV for ML / deep learning tasks. If we dont have our transform we may produce incorrect predictions, thumbnails and the like.

See https://github.com/PyAV-Org/PyAV/issues/570 for an earlier discussion

## Existing FFmpeg API

[av_display_rotation_get](https://ffmpeg.org/doxygen/trunk/group__lavu__video__display.html#gaac2ea94d3f66496c758349450b5b0217)

## Expected PyAV API

being able to introspect a video track and extract it rotation would add a class method to `av.video.stream.VideoStream`

something like `getDisplayRotation()` or perhaps a more pythonic `get_display_rotation` which would return a `np.array` containing the rotation matrix.

Should a track not have a preferred transform, I see two possible solutions. Return `None` or return an `identity matrix`

Other media processing apis take a similar approach, for example, `AVFoundation` which is Apples media processing api exposes a `prefferedTransform` property on a asset track (equivalent of a stream in ffmpeg parlance)

https://developer.apple.com/documentation/avfoundation/avpartialasyncproperty/3816142-preferredtransform

Thank you for your consideration, and apologies if I missed anything in the existing apI. Time permitting I may open a PR for this feature.

lgeiger:
This adds a `VideoFrame.rotation` property which uses [`av_display_rotation_get`](https://ffmpeg.org/doxygen/trunk/group__lavu__video__display.html#gaac2ea94d3f66496c758349450b5b0217) to read the rotation value from `DISPLAYMATRIX`.

~~This exposes a wrapper around [`av_display_rotation_get`](https://ffmpeg.org/doxygen/trunk/group__lavu__video__display.html#gaac2ea94d3f66496c758349450b5b0217) allowing users to read the rotation value from the `DISPLAYMATRIX`.~~

~~Not sure if this is the most usable API but at least this brings back the ability to read display rotation information.~~

Fixes #1047
Closes #1629


lgeiger:
> Not sure if this is the most usable API but at least this brings back the ability to read display rotation information.

A more user friendly API would be #1676 but will be a breaking/behaviour change for some users.

WyattBlue:
This API seems weird to me
Would `frame.rotation` be better?

hmaarrfk:
> This API seems weird to me
> Would frame.rotation be better?

My understanding is that the rotation is now part of sidechannel data in MP4 streams.

lgeiger:
> My understanding is that the rotation is now part of sidechannel data in MP4 streams.

If I understand the FFmpeg documentation correctly, the rotation information is only part of the [AVPacket side data](https://ffmpeg.org/doxygen/trunk/structAVPacket.html#ac55bfef91c33f02704ba76518d0f294c) (which PyAV doesn't expose yet) or part of the [AVFrame side data](https://ffmpeg.org/doxygen/trunk/structAVFrame.html#a44d40e03fe22a0511c9157dab22143ee).

I think `frame.rotation` would make sense not sure if there is a performance impact to reading the rotation for every frame. Maybe it would be more performant to have `packet.rotation` instead or in addition? Not sure whether this data can actually change from frame to frame or packet to packet in which case one could do some caching.

lgeiger:
> Would frame.rotation be better?

Done in 17b0b96991dc2b9b749ce76b2468c47f128767a0

The mypy failure in `tests/test_audioframe.py` seems unrelated.

PR #1698, opened 2024-12-29
Basic test for svtav1 encoder

JoeSchiff:
Test for https://github.com/PyAV-Org/pyav-ffmpeg/pull/139


WyattBlue:
This does not need a test

PR #1712, opened 2025-01-15
Support grayf32le and gbrapf32le in  numpy convertion.

robinechuca:
This pull request enables all 32-bit pixel formats to be supported.


robinechuca:
Thank you for the reactivity!

PR #1725, opened 2025-01-21
Fix backward compatibility of pull request 1722.

robinechuca:
Same as 1722, keeping classical python set.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,7 @@ Features
 
 - Add hardware decoding by :gh-user:`matthewlai` and :gh-user:`WyattBlue` in (:pr:`1685`).
 - Add ``VideoFrame.rotation`` by :gh-user:`lgeiger` in (:pr:`1675`).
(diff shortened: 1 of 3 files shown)


WyattBlue:
The underlying principle for `to_ndarray()` is that `bytes(to_ndarray(frame))` should be the same as the ffmpeg cli when returning the pix_fmt with `-c:v rawvideo`.

While I haven't checked this, I'm 90% sure that means putting the channel first in `yuv444p16`. I would like for the channel to be first in all relevant pix_fmts if the former condition holds. I'm okay with a `channel_last` parameter, but quite what you're doing.

Also more little things:
1. Don't put annotations in `*.pyx`, put then in the `*.pyi` files
2. Learn how to use feature branches in git so I don't have to "Squash and Merge" for your PRs in the future
3. Let's circle back to the `grb` vs `rgb` thing when I'm happy with this PR, leave it out of this PR for now

robinechuca:
Ha ok, I understand the logic now! This should be written in the `to_ndarray` documentation.
2. You mean you would like I work an an other branch than "main"?

PR #1737, opened 2025-01-29
Allow user to select audio codec when opening an audio device

z-khan:
This PR relates to #702 and #816
This enhancement lets user to specify any PCM audio codec to be used instead of relying on FFMPEG default (PCM_S16LE) when opening an audio device. Other than user preference, this is needed because some audio devices will not open with the default 16-bit codec e.g. some Focusrite Scarlett audio devices.

Usage:
```
import av
av.open(file="alsa_device_name", format="alsa:pcm_s32le")
```

Credits to @Kazhuu, @WyattBlue


Kazhuu:
Nice to see this getting in. I'm still in need for this feature actually :+1:

PR #1778, opened 2025-02-16
Stop publishing sdist

WyattBlue:
sdist takes ~4MB of space, is neglected by PyPA, and is a bad experience for both the developer and the user. PyAV does not have universal support for every arch, and removing source builds in PyPI does not change that.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,24 +4,6 @@ on:
     types: [published]
   workflow_dispatch:
 jobs:
-  package-source:
-    runs-on: ubuntu-24.04
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.13"
-      - name: Build source package
-        run: |
-          pip install setuptools cython
-          python scripts/fetch-vendor.py --config-file scripts/ffmpeg-7.1.json /tmp/vendor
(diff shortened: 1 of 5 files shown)


jlaine:
Please roll this back, this is really dodgy. Some users have an explicit policy of not trusting binary wheels and they should absolutely be able to "pip install"!

hmaarrfk:
Hmm, looking through recent changes and flagged this one as suspicious too.

We use the pypi sources for conda-forge because often, different packages will have "generated files" that get uploaded to pypi but are not included in the github tags.

https://github.com/conda-forge/av-feedstock/blob/main/recipe/meta.yaml#L8

I'm also suspicious of the claim:

> is neglected by PyPA,

 People can in fact use the source, and this was recommended for exotic platforms:

image

It would be good if you gave some time for downstream to comment on these changes.

I've been quite active and I would have been able to give give this feedback before you pressed merge...

WyattBlue:
@hmaarrfk I say neglected because `MANIFEST.in` is such a goofy format now that `pyproject.toml` exists.

Anyway, I'll let @jlaine release `14.2.0` when he feels everything is in order.

WyattBlue:
> Anyway, I'll let @jlaine release 14.2.0 when he feels everything is in order.

Retracted. jlaine is not talking to me at all. I've revered all objectionable changed, so I'm releasing now.

PR #1787, opened 2025-02-20
Remove outdated installion section

WyattBlue:


jlaine:
Why did you remove this documentation? We don't provide wheels for every single platform out there, so it was useful to have the requirements documented, no?

PR #1801, opened 2025-02-27
fix examples in audio atempo

Mddct:


WyattBlue:
The original is correct.

You can verify this with `rg --vimgrep "\).configure" tests`

PR #1816, opened 2025-03-12
Add getter/setter for global_header

daveisfera:
If there's not a better option, then this is the fix for #1814 (creating this so I can test a fully integrated build)

### DIFF
--- a/av/codec/context.pyx
+++ b/av/codec/context.pyx
@@ -121,6 +121,22 @@ cdef class CodecContext:
         self.ptr.flags = value
 
     @property
+    def global_header(self):
+        """
+        Use global header.
+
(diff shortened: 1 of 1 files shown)


WyattBlue:
This is not necessary, you can already read/write ptr.flags:

```python
from av.codec.context import Flags

video.codec_context.flags |= Flags.global_header
```

Even if this is a common operator, your PR would need proper type hints and a basic test.

daveisfera:
Awesome! This is what I was looking for and couldn't figure out.

Writing down how to clear the flag in case other stumble on this:
```
video.codec_context.flags &= ~Flags.global_header
 ```

PR #1835, opened 2025-03-26
Enable libx265 for ffmpeg in build-deps

tsnl:
When building `FFmpeg` as a dependency, the current build does not include `libx265`. 

It appears that release `PyAV` versions do support `libx265`. I'm guessing this change only affects development or those who build from source.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -57,13 +57,13 @@ jobs:
           ubuntu-24.04)
             sudo apt-get update
             sudo apt-get install autoconf automake build-essential cmake \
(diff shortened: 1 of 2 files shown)


WyattBlue:
This PR is bad because it does not think through the logical conclusions of its own premise. If adding x265 is clearly good, why not add svt-av1, aom, etc. Or even include fdk-aac with `--enable-nonfree`. The reason, besides legal issues, is that it just makes the CI take more storage and time than necessary. 

Unless you really need to test some feature of an encoder, there's no reason to include it in this script.

PR #1837, opened 2025-03-28
Use commands that work on macOS

daveisfera:


daveisfera:
The failed run made me realize that this isn't a macOS problem but a `zsh` vs `bash` problem, so I'll adjust the check

PR #1856, opened 2025-04-05
Provide binary wheels for musl-based Linux distros such as Alpine

jlaine:


jlaine:
I removed `yum install -y alsa-lib libxcb`:
- It's not needed
- It doesn't work for `musllinux` builds, on Alpine the package manager is `apk` not `yum`

PR #1880, opened 2025-04-29
Expose the Duration Field for Frames

Queuecumber:
This PR exposes the [duration](https://ffmpeg.org/doxygen/trunk/structAVFrame.html#ab90bd20a44c79c144045916b00c50e0d ) field on frame objects as the time that a given frame should be displayed for in `time_base` units. 

This can simplify certain decode loops which are searching for specific frames in time, e.g.: 

```
for frame in container.decode(video=0):
  if (frame.pts <= target_time_pts <= frame.pts + frame.duration):
    break # Found it
```

In this example the frame might not occur *at* the target pts so we need to use duration to check if the frame would be onscreen at the target time.

Currently this can be done by computing an average frame duration using the average frame rate (which is error prone) or trying to synchronize to the packets duration field.

### DIFF
--- a/tests/test_videoframe.py
+++ b/tests/test_videoframe.py
@@ -49,6 +49,21 @@ def test_opaque() -> None:
                 assert type(frame.opaque) is tuple and len(frame.opaque) == 2
 
 
+def test_frame_duration_matches_packet() -> None:
+    with av.open(fate_suite("h264/interlaced_crop.mp4")) as container:
+        packet_durations = [
+            (p.pts, p.duration) for p in container.demux() if p.pts is not None
+        ]
+        packet_durations.sort(key=lambda x: x[0])
+
+    with av.open(fate_suite("h264/interlaced_crop.mp4")) as container:
+        frame_durations = [(f.pts, f.duration) for f in container.decode(video=0)]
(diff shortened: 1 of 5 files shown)


tests/test_videoframe.py
```diff
@@ -49,6 +49,24 @@ def test_opaque() -> None:
                 assert type(frame.opaque) is tuple and len(frame.opaque) == 2
 
 
+def test_frame_duration_matches_packet() -> None:
+    with av.open(fate_suite("h264/interlaced_crop.mp4")) as container:
```
Queuecumber, line 53: Question for maintainers: is there a more interesting video we could put here? This one seems to have `duration=1` for all packets

WyattBlue: Something with VFR would do. This test is fine though, considering how annoying, at times, finding stuff in FATE can be.

WyattBlue: This can be added in another PR

Queuecumber: OK thanks, will look into the FATE videos and find one that works for a followup PR

Queuecumber:
Thanks for the fast turnaround on this

PR #1913, opened 2025-06-07
Fix memory error for AudioFifo properties

zzjjbb:
Validate self.ptr before access: uninitialized properties should return None rather than crash the Python interpreter


zzjjbb:
Looks like I'm using a very old version...

zzjjbb:
I don't know why `format`, `layout`, and `sample_rate` are looked up in the same way, but `AudioFifo().sample_rate` will raise an `AttributeError`, while `AudioFifo().format` and `AudioFifo().layout` will crash the interpreter. Is there any magic caused by the docstrings in Cython?

WyattBlue:
I think setting `sample_rate` to `0` and raising `AttributeError`s is better than using `None`. Otherwise, this type widening would be a "breaking change"

PR #1944, opened 2025-07-03
Allow user to set its own PYAV_TESTDATA_DIR in tests/common.py

nicolas-graves:
There was no user-friendly way to properly set the PYAV_TESTDATA_DIR, because `os.environ["PYAV_TESTDATA_DIR"] = asset()` overrides the previous value. Use `os.environ.setdefault("PYAV_TESTDATA_DIR", asset())` instead.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -66,7 +66,7 @@ def asset(*args: str) -> str:
 
 
 # Store all of the sample data here.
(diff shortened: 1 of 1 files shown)


WyattBlue:
I do not want to associate with you.

PR #1964, opened 2025-07-22
Add set_image for updating the image content of an existing frame

daveisfera:
When updating the content of a video, this is more efficient because the existing frame can be used


daveisfera:
This isn't useful like I had hoped and will require some additional work to be valuable.
The current `to_image` and `from_image` functions have an implied use of `RGB` format, so it's not a drop in replacement to use the existing frame, because calling `set_image` on it is trying to copy into a `YUV` image in the tests I've done and the current `copy_bytes_to_plane` appears to be assuming `RGB` so it doesn't work.

Issue #1987, opened 2025-08-21
Seek succeeds but any frame read is all zeros [revisited]


tikuma-lsuhsc:
This issue was originally discussed in Issue #349 in 2018 (moved to Discussion as #928), but it wasn't resolved then (as far as I can tell). I just came across this very issue and may be able to shed additional light.

Here is a minimal example (tested with v15.0.0 running on Python v3.13.3 in Linux):

```python
import av

seek_first = False
with av.open(avifile) as fmt:
    vst = fmt.streams.video[0]
    if seek_first:
        fmt.seek(0, stream=vst, backward=True, any_frame=False)
    frame = next(fmt.decode(video=0))
    data = frame.to_ndarray(format="rgb24")
    assert np.any(data!=0) # errors out only if seek_first=True
```

The assertion appears to fail only with my AVI file (uncompressed video from highspeed video camera). If I use an MP4 file, it worked properly. Also, every frame of this AVI file is a key frame. I cannot share my video file but this info may help debugging the issue.

Thanks, and Thanks for the awesome package

P.S., I'd be happy to move this post under #928 if appropriate, but it does appear to be a real issue.

tikuma-lsuhsc:
A follow up on generating a video file with this issue:

```
ffmpeg -f lavfi -i testsrc=duration=5.3:size=qcif:rate=10 -vcodec rawvideo -pix_fmt pal8 test_video.avi
```

The main culprit seems to be the rawvideo PAL8 format.

tikuma-lsuhsc:
This issue is addressable with post-v15.1 releases with the PR #2003.

The cause of this behavior is that `InputContainer.seek()` clears the codec context's palette. This is FFmpeg behavior and not PyAV behavior. In order to restore the palette, we must (1) intercept the palette(s) which are attached to packets as a sidedata, (2) make a copy of them, and (3) attach the appropriate one to the first packet obtained after a seek.

Here is an example:

```python

with av.open(avifile) as fmt:

    vst = fmt.streams.video[0]

    # get the palette from the first packet
    pkt = next(fmt.demux(vst))
    assert pkt.has_sidedata("palette") # expect the palette to be defined in the first packet
    palette = pkt.get_sidedata('palette')

    # seek (rewind to the beginning, which does not re-retrieve the palette)
    fmt.seek(0, stream=vst, backward=True, any_frame=False)

    # manually set the palette to the first post-seek packet so codec can re-initialize its palette
[...]

PR #1989, opened 2025-08-24
fixed seek operation when not at eof

tikuma-lsuhsc:
This is a PR to resolve Issue #1987 ("Seek succeeds but any frame read is all zeros [revisited]") that I posted earlier.

The problem was that the said AVI format and configuration cannot have the codec buffer flushed if not at the end of the file. I studied FFmpeg's [`ffplay.c`](https://github.com/FFmpeg/FFmpeg/blob/master/fftools/ffplay.c) for their handling of during-playback seeks, and the only difference is that they do not flush the codec buffers. 

If we simply remove: https://github.com/PyAV-Org/PyAV/blob/0b85c60defbea05ee8118a0605f7c31254589a0b/av/container/input.pyx#L279
It fails the `test_decode_half` and `test_stream_seek` because seeking after reaching the EOF does not reinstate the codecs. To get the codecs output of the EOF, we must call `flush_buffers()`. 

To satisfy these conditions, this PR adds an `eof` flag to `InputContainer` and adopted the EOF checking mechanism from ffplay: https://github.com/FFmpeg/FFmpeg/blob/673f28b6cb6e20f5a695a90d144a6158bdf987fe/fftools/ffplay.c#L3098

P.S., I've also added `InputContainer.seek2()` method which wraps `avformat_seek_file()` to my fork while I was investigating this issue. I'd be happy to add it to this PR (or as another PR) if you find it a reasonable addition to the class.

### DIFF
--- a/av/container/input.pxd
+++ b/av/container/input.pxd
@@ -6,4 +6,6 @@ from av.stream cimport Stream
 
 cdef class InputContainer(Container):
 
+    cdef bint eof
+
     cdef flush_buffers(self)
--- a/av/container/input.pyx
+++ b/av/container/input.pyx
@@ -173,6 +173,7 @@ cdef class InputContainer(Container):
                         ret = lib.av_read_frame(self.ptr, packet.ptr)
                     self.err_check(ret)
                 except EOFError:
+                    self.eof = True
                     break
 
                 if include_stream[packet.ptr.stream_index]:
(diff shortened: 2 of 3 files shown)


tikuma-lsuhsc:
Sorry, I pulled the gun a little too early. This still doesn't fix the problem. I'll investigate further.

PR #2013, opened 2025-09-17
Prevent segfault on data stream template

dotysan:
I was wrongly attempting to copy a data stream using add_stream_from_template() and got a segfault instead of a useful response. This patch/test resolves.

The segfault occurs on assigning `None.codec` or `None.name` to `codec_obj: Codec`.

### DIFF
--- a/tests/test_streams.py
+++ b/tests/test_streams.py
@@ -1,5 +1,6 @@
 import os
 from fractions import Fraction
+from typing import Any, cast
 
 import pytest
 
@@ -147,7 +148,6 @@ class TestStreams:
 
         container.close()
 
-''' Thought this was helpful, but it now breaks add_stream_from_template() -> _StreamT
     def test_data_stream_from_template(self) -> None:
         """Test that adding a data stream from a template raises ValueError."""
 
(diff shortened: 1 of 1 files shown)


dotysan:
OK, clearly I'm going about this wrong.
```
mypy av tests
tests/test_streams.py:160: error: Value of type variable "_StreamT" of "add_stream_from_template" of "OutputContainer" cannot be "DataStream"  [type-var]
Found 1 error in 1 file (checked 88 source files)
make: *** [Makefile:34: lint] Error 1
```

Why doesn't this in output.pyi prevent the segfault in the first place?!?
```py
class OutputContainer(Container):
    ...
    def add_stream_from_template(
        self, template: _StreamT, opaque: bool | None = None, **kwargs
    ) -> _StreamT: ...
```

WyattBlue:
`.pyi` files are just type stubs. They help mypy but don't affect behavior

dotysan:
Is there something I can do to get that Windoze smoke test to pass?

WyattBlue:
Windows failing seems unrelated. If you can fix this, make a separate PR

PR #2014, opened 2025-09-17
prevent rsync error on missing subdir

dotysan:
trivial fix


dotysan:
Is there something I can do to get that Windoze smoke test to pass?

PR #2015, opened 2025-09-18
Update two type stubs for the use of `None`

velsinki:
Please see the individual commits for the reasoning.

I found these issues using Mypy.


velsinki:
@WyattBlue, thanks for merging so fast 😅 

I had missed the typing problem in the tests though. Maybe should add some assertions there to satisfy Mypy?

WyattBlue:
Yes, that would be great

velsinki:
Alright, see #2016 :+1:

Issue #2148, opened 2026-02-03
Can I connect GPU frames to another interface without copying?


caffeinism:
It would be great if everything could be done in ffmpeg, but I discovered that overlay_cuda doesn't support transparent frames (in fact, I realized there are no frames with alpha channels supported on the GPU to begin with). I want to decode video using nvdec (cuvid) in pyav, copy it to PyTorch for processing, then move it back to pyav for encoding with nvenc.

However, I see no signs that anyone has attempted this, so I'm wondering if it's actually possible. Is there anyone who might be able to help?

caffeinism:
... Looking at the logs, it appears pyav immediately transfers decoded frames to the CPU and frees the original CUDA memory. Is it not possible to process?

caffeinism:
https://github.com/PyAV-Org/PyAV/discussions/1893#discussion-8325048

calllivecn:
I have the same need.

PR #2155, opened 2026-02-04
Preserving hardware memory during cuvid decoding, exporting/importing via dlpack.

caffeinism:
https://github.com/PyAV-Org/PyAV/issues/2148

Hello? I'm a user with limited knowledge of libav, dlpack, and cython. However, recognizing this as a necessary feature, I drafted this with the help of an LLM.

### Motivation
If an application decodes video, performs GPU operations, and then re-encodes it, PyAV currently incurs a significant amount of memcopy. (GPU (cuvid) -> CPU (PyAV) -> GPU (Torch, etc.) -> CPU (PyAV) -> GPU (nvenc)) However, if we could export frames decoded by cuvid to dlpack while keeping them on the GPU, we wouldn't need to move the frames to CPU memory.

I passed all existing tests, but with such extensive modifications, it seems difficult for a beginner like me to catch every single detail. However, since most changes involve adding features rather than modifying existing ones, I hope this PR serves as a good starting point.

### Usage example
```
import av
from av.codec.hwaccel import HWAccel
import torch

hwaccel = HWAccel(
    device_type="cuda",
    device=0,
    allow_software_fallback=False,
    output_format="hw", # preserve hw memory
)

# decode using cuvid
with av.open(from_video_filename, "r", hwaccel=hwaccel) as c:
    frame = next(c.decode(video=0))
    y = torch.from_dlpack(frame.planes[0]) # device(type='cuda', index=0), torch.uint8, torch.Size([H, W])
    uv = torch.from_dlpack(frame.planes[1]) # device(type='cuda', index=0), torch.uint8, torch.Size([H/2, W/2])

f = av.VideoFrame.from_dlpack(((y*0.5).to(torch.uint8), uv)) # some operation

with av.open(to_video_filename, "w") as c:
    s = c.add_stream("h264_nvenc", rate=24) # encode using nvenc
    for it in s.encode(f):
        c.mux(it)
    for it in s.encode(None):
        c.mux(it)

### CODE REVIEW

**Comment by caffeinism** [comment]
@WyattBlue If I add tests, will it work fine even if it only runs on a CUDA machine? I don't think it will work in the GitHub workflow.

**Comment by WyattBlue** [comment]
You need to test the interface. For example, hw_format does not have an `pyi` interface, and writing a test would catch that fact.

**Comment by WyattBlue** [comment]
`av/hwcontext.pxd‎` should be merged with `include/avutil`. `*.pxd` files should otherwise not be free radicals, i.e., they should have a corresponding real `.py` file.

**Comment by caffeinism** [comment]
> You need to test the interface. For example, hw_format does not have an `pyi` interface, and writing a test would catch that fact.

Could you please explain it in a bit more detail?

> `av/hwcontext.pxd‎` should be merged with `include/avutil`. `*.pxd` files should otherwise not be free radicals, i.e., they should have a corresponding real `.py` file.

In this case, how should dlpack.pxd be handled? Should this also be moved to the include directory?

**Comment by caffeinism** [comment]
@WyattBlue Could you take a look at the last commit section? I modified the buffer creation logic for frames generated by VideoFrame(), VideoFrame.from_ndarray(), and VideoFrame.reformat to support dlpack.

**Comment by WyattBlue** [comment]
```
  _cuda_device_ctx_cache = {}
  _cuda_frames_ctx_cache = {}
```
  These grow indefinitely with no cleanup mechanism. For long-running applications using many different frame sizes, this could lead to memory growth. The global caches in frame.py and the registry in _hwdevice_registry.py are not thread-safe. Worth documenting or addressing
  
  Redundant None check (av/video/frame.py:596-597)
  if primary_ctx is None:
      primary_ctx = True
  The function signature already has primary_ctx: bool = True, so this check is
  unreachable.
  
  Duplicate validation (av/video/frame.py:498-500)
  if dev0 != dev1:
      raise ValueError("plane tensors must be on the same CUDA device")
  if dev_type0 == kDLCUDA:
      if dev0 != dev1:  # Redundant - already checked above
          raise ValueError("plane tensors must be on the same CUDA device")
  The second check at line 499-500 is redundant since dev0 != dev1 is already
  checked at line 496-497.
  
  I'm not sure why we're import `PyObject`, imports that aren't used should be removed.

**Comment by caffeinism** [comment]
> ```
>   _cuda_device_ctx_cache = {}
>   _cuda_frames_ctx_cache = {}
> ```
> 
> These grow indefinitely with no cleanup mechanism. For long-running applications using many different frame sizes, this could lead to memory growth. The global caches in frame.py and the registry in _hwdevice_registry.py are not thread-safe. Worth documenting or addressing

This part is complex due to my limited implementation skills, so I left the global cache in place reluctantly. I'll give it more thought.

> Redundant None check (av/video/frame.py:596-597) if primary_ctx is None: primary_ctx = True The function signature already has primary_ctx: bool = True, so this check is unreachable.

There was some deliberation about whether to set the default value to bool | None = None or bool = True. I will delete that paragraph.

> Duplicate validation (av/video/frame.py:498-500) if dev0 != dev1: raise ValueError("plane tensors must be on the same CUDA device") if dev_type0 == kDLCUDA: if dev0 != dev1: # Redundant - already checked above raise ValueError("plane tensors must be on the same CUDA device") The second check at line 499-500 is redundant since dev0 != dev1 is already checked at line 496-497.

I didn't notice it during my review.

> I'm not sure why we're import `PyObject`, imports that aren't used should be removed.

I realized that PyObject* doesn't have a reference count and replaced it with object, but I forgot to remove the import. I will check this part as well.

File: av/video/codeccontext.py [hunk 1]
```diff
@@ -127,9 +127,7 @@ def _transfer_hwframe(self, frame: Frame):
             # need to transfer.
             return frame
 
-        if self.hwaccel_ctx.output_format == "hw":
-            vframe = cython.cast(VideoFrame, frame)
-            vframe._device_id = int(getattr(self.hwaccel_ctx, "device_id", 0))
```
caffeinism: Since the structure has changed to retrieve the device number from frame.device_id instead of fetching it from the global cache, I believe issues may arise when the change occurs if device_id!=0.

caffeinism:
Thanks!

If anyone happened to come here via the issue, torch's default `from_dlpack` implementation seems to use `stream != None` by default, so you can use something like `torch.from_dlpack(frame.planes[0].__dlpack__())`.

Issue #2161, opened 2026-02-10
Segfault when input contains some corrupted streams


LasSino:
Code like this will cause segfault for some inputs with corrupted or invalid streams:
```python
import av
container = av.open("input.m2ts", "r")  # for certain input files
stream = container.streams[1] # select the corrupted stream
print(stream.type) # segfault here
```

This happens when the stream is corrupted so a general `Stream` rather than specific stream like `VideoStream` is created. So the `type` property invokes:
```python
# av/stream.py
@cython.cclass
class Stream:
    ...

    @property
    def type(self):
        """
        The type of the stream.

        :type: Literal["audio", "video", "subtitle", "data", "attachment"]
        """
        return lib.av_get_media_type_string(self.ptr.codecpar.codec_type)
```

The underlying `av_get_media_type_string` function may return `NULL` for unknown `AVMediaType`:
```C
// FFmpeg/libavutil/utils.c
const char *av_get_media_type_string(enum AVMediaType media_type)
{
    switch (media_type) {
    case AVMEDIA_TYPE_VIDEO:      return "video";
    case AVMEDIA_TYPE_AUDIO:      return "audio";
    case AVMEDIA_TYPE_DATA:       return "data";
    case AVMEDIA_TYPE_SUBTITLE:   return "subtitle";
    case AVMEDIA_TYPE_ATTACHMENT: return "attachment";
    default:                      return NULL;
    }
}
```

When `NULL` is returned, a segfault will be triggered since Cython converts the returned `const char *` to a Python string without checking for `NULL`.

PR #2162, opened 2026-02-10
Fix: segfault when av_get_media_type_string returns NULL

LasSino:
Resolves: #2161.
- Adds `NULL` check for function return.
- Updates function signature in `.pyi` file.

*Btw: An empty string is returned when the function returns `NULL` in the fix. I'm not quite sure about that. Maybe an "unknown" is preferred?*


WyattBlue:
Superseded by #2163

PR #2167, opened 2026-02-19
Remove cython workaround from input container

lgeiger:
it looks like the workaround is not needed anymore in recent versions of Cython: https://github.com/cython/cython/issues/2166

### DIFF
--- a/av/container/input.py
+++ b/av/container/input.py
@@ -151,13 +151,7 @@ class InputContainer(Container):
         """
         self._assert_open()
 
-        # For whatever reason, Cython does not like us directly passing kwargs
-        # from one method to another. Without kwargs, it ends up passing a
-        # NULL reference, which segfaults. So we force it to do something with it.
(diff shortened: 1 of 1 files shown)


av/container/input.py
```diff
@@ -151,13 +151,7 @@ def demux(self, *args, **kwargs):
...
         self._assert_open()
 
-        # For whatever reason, Cython does not like us directly passing kwargs
-        # from one method to another. Without kwargs, it ends up passing a
-        # NULL reference, which segfaults. So we force it to do something with it.
-        # This is a bug in Cython; see https://github.com/cython/cython/issues/2166
-        id(kwargs)
-
-        streams = self.streams.get(*args, **kwargs)
+        streams: list[Stream] = self.streams.get(*args, **kwargs)
```
lgeiger: Adding some types helps Cython specialise and generate less code

PR #2236, opened 2026-04-18
Use `layout_compare()` for `AudioLayout.__eq__()`

WyattBlue:
And use AVCodecContext.frame_num instead of keeping track ourselves.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,13 +31,16 @@ v17.0.1 (next)
 Fixes:
 
 - A cdivision decorator for faster division.
+- Adjust tests so they work with 8.1 by :gh-user:`strophy`.
 - Make VideoFormat.components lazy by :gh-user:`WyattBlue`.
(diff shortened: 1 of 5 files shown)


Dobatymo:
@WyattBlue Hi! How can this be part of a patch release if it changes the public api by removing `encoded_frame_count`? My code is relying on that. How could I replace it? It would be nice if the docs could contain a note about this and how to adjust to it.

WyattBlue:
If some property is not exercised in the test suite, it's not part of the public api. This api wasn't documented either, it was just internal plumbing that was made read-only instead of private for some reason.

If your code does rely on this, just use v16.1.0. Consider this change a post release for v17.

Dobatymo:
@WyattBlue Thank you for your response. I assumed it was public since it was documented here https://pyav.basswood-io.com/docs/16.1/api/video.html#av.video.codeccontext.VideoCodecContext.encoded_frame_count

PR #2294, opened 2026-06-07
Add out= and VideoPlane.to_ndarray for preallocated/partial reads

hmaarrfk:
I've been wanting to enable this for quite some time since it would help me in a few "performance" constrained scenerios.

It doesn't use the frame allocation callbacks from ffmpeg, but historically, i've found that things like that are seldom tested (by the C-side projects), so i wanted to get an "out" parameter before diving into that

<details><summary>Claude's draft</summary>

Two additions that let callers read frame data into their own buffers and read a single plane without assembling the whole frame:

- `VideoFrame.to_ndarray(out=None)`: when `out` is given, the conversion writes into that preallocated, C-contiguous array (validated for shape and dtype up front) and returns it, avoiding a per-frame allocation. The flattened planar formats (yuv420p/yuvj420p/yuv422p/nv12) copy each plane straight into its slice of `out` instead of building an intermediate hstack. Not supported for `pal8` (which returns a tuple).

- `VideoPlane.to_ndarray(out=None)`: returns a single, single-component plane as a 2D `(height, width)` array with line padding removed -- for example the luma (Y) plane of a planar YUV frame as a grayscale image, without ever touching the chroma planes. Returns a zero-copy view when `out` is None, or copies into `out`. Packed/interleaved planes are rejected with a clear error.

Includes tests for both (round-trips across formats, validation errors, negative line_size, and luma-plane equivalence) and a CHANGELOG entry.

Resume this Claude session:
```
claude --resume 50dbe019-cdd6-48e4-a346-6976cd007096
```
</details>


av/video/frame.py
```diff
@@ -784,34 +807,78 @@ def to_ndarray(self, channel_last=False, **kwargs):
...
+                array[:, :, i] = byteswap_array(
+                    useful_array(planes[p_idx], itemsize, dtype), big_endian
+                )
+
+            if not transpose:
+                return array
+            if out is None:
+                return np.moveaxis(array, 2, 0)
+            out[...] = np.moveaxis(array, 2, 0)
+            return out
```
hmaarrfk, line 846: this code is not using the pre-allocated array correct. it always create array regardless.

instead it should branch early based on the value of out and write the copy algorithm in two different ways for both cases.

av/video/frame.py
```diff
@@ -823,10 +890,15 @@ def to_ndarray(self, channel_last=False, **kwargs):
...
+            array = np.stack([y, u_full, v_full], axis=2 if channel_last else 0)
+            if out is None:
+                return array
+            _check_out(out, array.shape, "uint16")
+            out[...] = array
+            return out
         if format_name == "pal8":
+            if out is not None:
+                raise ValueError("out is not supported for the pal8 format")
             image = useful_array(planes[0])
```
hmaarrfk, line 899: you should support out for pal8 here, the image should go into it.

hmaarrfk:
i'm not sure i really want this. it is "nice" but there has to be a better way sorry for the PR noise

PR #2323, opened 2026-06-22
Apply reformat shortcut to hardware frames too

CarlosRDomin:
This is a small follow-up to #2321 

When `VideoFrame.reformat` is called on a hardware frame, PyAV first downloads it into a software frame. If that downloaded frame already matches the requested output format, size, colorspace and color range, we can return it directly instead of passing it through swscale. This avoids unnecessary e.g. `nv12 -> nv12` swscale conversions after hardware download, which besides being inefficient actually modify pixel values even though no format/size conversion is requested. Without my change to `reformatter.py`, the test I introduce in `test_decode.py` fails on the latest PyAV (both `main` and 17.1.0).

The included test compares:
 - a hwaccel frame decoded with `is_hw_owned=False`, which PyAV downloads directly, and
 - a hwaccel frame decoded with `is_hw_owned=True` and then reformatted to the same software format.

With the shortcut, both paths preserve identical pixel data and the test passes.

### DIFF
--- a/av/video/reformatter.py
+++ b/av/video/reformatter.py
@@ -291,6 +291,19 @@ class VideoReformatter:
             frame = frame_sw
             src_format = cython.cast(lib.AVPixelFormat, frame.ptr.format)
 
+            # Check for shortcut again, in case dst_format matches the downloaded frame's sw_format
+            if (
+                dst_format == src_format
+                and width == frame.ptr.width
+                and height == frame.ptr.height
+                and dst_colorspace == src_colorspace
+                and src_color_range == dst_color_range
+                and not set_dst_color_trc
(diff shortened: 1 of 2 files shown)


CarlosRDomin:
Minor question @WyattBlue: instead of manually checking that the dimensions, format, colorspace/range match, should we instead call [`sws_is_noop`](https://ffmpeg.org/doxygen/trunk/group__libsws.html#ga50bb392cac2bf4e626e309b56ab76c32)?

WyattBlue:
That seems like a good idea

CarlosRDomin:
Cool, happy to give it a try this evening and submit a PR for your review :)

PR #2327, opened 2026-06-24
Leverage sws_is_noop for video reformat shortcuts

CarlosRDomin:
Following up on https://github.com/PyAV-Org/PyAV/pull/2323#issuecomment-4782076149, this PR replaces the manual video reformat no-op checks with FFmpeg’s `sws_is_noop`.

Hardware frames are downloaded before the no-op check so the comparison is performed against the software frame that would actually be passed to swscale, so we only check for the no-op shortcut once. The destination frame is prepared as a lightweight descriptor for `sws_is_noop`, and buffers are only allocated if `sws_scale_frame` is needed.

The change also simplifies color transfer/primaries handling by resolving `None` to the source frame values before `_reformat`, removing the extra `set_dst_color_*` booleans while preserving explicit `UNSPECIFIED` behavior.

### DIFF
--- a/av/video/reformatter.py
+++ b/av/video/reformatter.py
@@ -220,17 +220,15 @@ class VideoReformatter:
         )
         c_src_color_range = _resolve_enum_value(src_color_range, ColorRange, 0)
         c_dst_color_range = _resolve_enum_value(dst_color_range, ColorRange, 0)
-        c_dst_color_trc = _resolve_enum_value(dst_color_trc, ColorTrc, 0)
+        c_dst_color_trc = _resolve_enum_value(
+            dst_color_trc, ColorTrc, frame.ptr.color_trc
+        )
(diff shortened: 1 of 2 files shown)


av/video/reformatter.py
```diff
@@ -267,64 +259,54 @@ def _reformat(
         dst_color_range: cython.int,
         dst_color_trc: cython.int,
         dst_color_primaries: cython.int,
-        set_dst_color_trc: cython.bint,
-        set_dst_color_primaries: cython.bint,
         threads: cython.int,
     ):
         if frame.ptr.format < 0:
             raise ValueError("Frame does not have format set.")
```
CarlosRDomin: FYI not sure if this is needed anymore (I left it just in case, but maybe sws is robust to it already?)

WyattBlue: I would just remove it

PR #2332, opened 2026-06-29
Allow configuring hardware encoder sw_format

CarlosRDomin:
Currently hardware encode setup derives `AVHWFramesContext.sw_format` from `AVCodecContext.pix_fmt`. If the user sets `pix_fmt="cuda"`, PyAV treats that as the hardware format and falls back to `nv12` for software format, with no public way to request a different software format such as `yuv420p`.

This PR adds a setter for `VideoCodecContext.sw_format` and makes `_setup_encode_hwframes()` prefer `AVCodecContext.sw_pix_fmt` when it is set, while preserving the existing `stream.pix_fmt = "nv12"` path and the current
`nv12` default.

The new `test_hardware_encode_honors_sw_format` test verifies that `pix_fmt="cuda"` plus `sw_format="yuv420p"` opens the encoder with a CUDA hardware frames context whose software format remains `yuv420p` instead of `nv12` (before this PR). Also tested on Mac with `videotoolbox` hwaccel with similar results.


tests/test_encode.py
```diff
@@ -569,3 +576,42 @@ def test_hardware_encode() -> None:
...
+        device_type=device_type, allow_software_fallback=False
+    )
+    container = av.open(io.BytesIO(), mode="w", format="mp4")
+    stream = container.add_stream(encoder, rate=30, hwaccel=hwaccel)
+    stream.width = 320
+    stream.height = 240
+    stream.pix_fmt = hw_format
+    stream.codec_context.sw_format = "yuv420p"
+
+    assert stream.codec_context.sw_format.name == "yuv420p"
```
CarlosRDomin, line 609: Note that this used to return None (`hw_frames_ctx` was NULL), now we allow getting (and setting) the value

tests/test_encode.py
```diff
@@ -569,3 +576,42 @@ def test_hardware_encode() -> None:
...
+
+    frame = VideoFrame(320, 240, "rgb24")
+    for packet in stream.encode(frame):
+        container.mux(packet)
+
+    assert stream.codec_context.pix_fmt == hw_format
+    assert stream.codec_context.sw_format.name == "yuv420p"
```
CarlosRDomin, line 617: Note that this used to return "nv12" (and would make the encoder use nv12 as the pixel format, with no way to control the value), now the sw_format property is respected

PR #2360, opened 2026-07-26
feat(nvenc): add support for explicit cuda streams

Yozer:
The performance gains probably depend on the specific pipeline.

In my case, I didn’t see any improvement with 1080p HEVC videos, but at 4K and 8K, the speed-up was around 13% and 25%, respectively (e2e wall time).

I haven’t looked into it too deeply, but my guess is that NVENC uses the default stream 0 and requires a full synchronization before encoding. At lower resolutions, that overhead doesn’t make much difference, but with 4K and 8K workloads, waiting for all the processing kernels to finish before encoding becomes more significant.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,7 @@ Features:
 - Add ``Packet.rescale_ts()`` to rescale packet PTS, DTS, and duration to a new ``AVRational`` time base by :gh-user:`WyattBlue`.
 - Support reusing the thread's current CUDA context via a ``current_ctx`` flag on ``CudaContext`` and ``VideoFrame.from_dlpack``, for interop with libraries like PyTorch that initialize CUDA first by :gh-user:`Yozer` (:pr:`2339`).
 - ``VideoFrame.from_dlpack`` no longer requires restating ``primary_ctx``/``current_ctx`` when passing an explicit ``cuda_context``; the flags are only validated when explicitly given by :gh-user:`WyattBlue`.
+- Support passing an explicit CUDA stream to FFmpeg CUDA operations, including NVENC input and output, via a ``cuda_stream`` parameter on ``CudaContext``; currently limited to logical CUDA device 0 by :gh-user:`Yozer` (:pr:`2360`).
 
 Fixes:
 
--- a/av/video/frame.pxd
+++ b/av/video/frame.pxd
@@ -1,5 +1,5 @@
 cimport libav as lib
-from libc.stdint cimport uint8_t
+from libc.stdint cimport uint8_t, uintptr_t
 
 from av.frame cimport Frame
 from av.video.format cimport VideoFormat
@@ -10,6 +10,7 @@ cdef class CudaContext:
     cdef readonly int device_id
     cdef readonly bint primary_ctx
     cdef readonly bint current_ctx
+    cdef readonly uintptr_t cuda_stream
     cdef lib.AVBufferRef* _device_ref
     cdef dict _frames_cache
     cdef lib.AVBufferRef* _get_device_ref(self)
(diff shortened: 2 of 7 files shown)


WyattBlue:
Found one blocking concern:

Avoid modifying an initialized FFmpeg device context — frame.py lines 89–105 (https://github.com/PyAV-Org/PyAV/blob/dbef919b14e5cd2ef3606b04faf5cc16fc8085f5/av/video/frame.py#L89-L105).

av_hwdevice_ctx_create() returns an initialized context, after which the PR writes directly to AVCUDADeviceContext.stream. FFmpeg explicitly states that hwctx must not be modified after av_hwdevice_ctx_init(). Although current NVENC happens to read the stream later during setup, this relies on behavior outside FFmpeg’s API contract. The stream should be supplied through a supported pre-initialization/creation path. FFmpeg API documentation (https://www.ffmpeg.org/doxygen/8.0/structAVHWDeviceContext.html)

Yozer:
You’re right. Sorry, I have little experience with ffmpeg av.
It should look better now. My pipeline still has perf gain.
I also updated the changelog.

WyattBlue:
The previous post-initialization mutation concern is resolved, but I found one remaining multi-GPU issue:

The secondary CUDA device context created in `CudaContext._get_device_ref()` is zero-initialized and copies only `cuda_ctx` and `stream` before calling `av_hwdevice_ctx_init()`:
https://github.com/PyAV-Org/PyAV/blob/0b8d386fc1e47ebfacad7ae62fe0e08be4f86ae9/av/video/frame.py#L116-L139

In FFmpeg 8.1.2, `av_hwdevice_ctx_init()` only loads the CUDA functions; it does not populate the private `cuda_device` field. The normal creation path sets that field from the requested device, and CUDA frame initialization later uses it to query texture alignment:
- https://github.com/FFmpeg/FFmpeg/blob/n8.1.2/libavutil/hwcontext_cuda.c#L320-L335
- https://github.com/FFmpeg/FFmpeg/blob/n8.1.2/libavutil/hwcontext_cuda.c#L445-L463
- https://github.com/FFmpeg/FFmpeg/blob/n8.1.2/libavutil/hwcontext_cuda.c#L150-L167

As written, `CudaContext(device_id > 0, cuda_stream=...)` therefore combines the requested GPU context with device-0 metadata, which can produce an incorrect frame-pool alignment.

Yozer:
Hmm, I wonder what would be best way to fix this one.
From what I can see the cuda device is hold in ffmpeg private struct - we could do hacky modifications it but I would like to avoid it.

The only solution I can see is to add a guard so it works for first gpu only. To be more specific: streams feature can be only used for device_id=0.
People could still in theory use different gpus via cuda visibility env variable.
Let me know @WyattBlue if you want me to add this guard.

The cleanest solution would be to improve ABI ffmpeg contract but I'm not sure if I want to go this route alone.

Yozer:
In theory I could also push owner CUcontext, call cuCtxGetDevice and compare that device’s CU_DEVICE_ATTRIBUTE_TEXTURE_ALIGNMENT with logical device 0. But I'm not sure if its any better.

//edit
and I don't even see pyav importing cuda libs so that's probably bad idea/too complex to implement

WyattBlue:
Yes, please add the guard. Reject cuda_stream when device_id != 0, with an error explaining that users can expose the desired physical GPU as logical device 0 through CUDA_VISIBLE_DEVICES.

I agree that modifying FFmpeg’s private struct or loading CUDA solely to reconstruct its private state would be too brittle. Supporting arbitrary device IDs cleanly should wait for an upstream FFmpeg API that accepts an external stream during device creation.

Yozer:
done

PR #2393, opened 2026-08-22
Expose the chroma location of a video frame

hmaarrfk:
As the title says.

There was to be a value VideoCodecContext, but no "per frame property"

This meerly exposes the ffmpeg property

<details><summary>Claude's draft</summary>

`VideoCodecContext.chroma_sample_location` already wrapped `AVCodecContext.chroma_sample_location`, but the frame side of the same field had no wrapper, so the chroma siting a decoder actually reported per frame could not be read at all — only the container's declaration.

Add `VideoFrame.chroma_location`, wrapping `AVFrame.chroma_location`. Each property mirrors its C field name, which FFmpeg spells differently on the two structs.

Add a `ChromaLocation` enum alongside `ColorRange`, `ColorTrc`, and `ColorPrimaries` in `av.video.reformatter`, naming the `AVCHROMA_LOC_*` values, and document it in the video API enum table. `AVChromaLocation` was already declared in `include/avutil.pxd`; `AVFrame.chroma_location` was not, and is now.

`VideoStream` proxies its codec context, so its stub gains the `chroma_sample_location` it was already forwarding.

Tests cover the frame property round trip, the codec context property, and an ffv1/matroska encode-decode round trip that carries TOPLEFT through to the decoded frames.

Resume this Claude session:
```
cd /home/mark/git/PyAV
claude --resume f640d6be-5fdb-42cd-b340-64abdde5eadb
```
</details>


hmaarrfk:
thanks.

Issue #2276, opened 2026-06-04
Logging causes excessive lock contention


oakaigh:
The default logging makes pyav almost unusable for thread-parallelized workloads. Below is a trace obtained by `py-spy`:

```
$ py-spy top --pid 2875341 --native --full-filenames
  %Own   %Total  OwnTime  TotalTime  Function (filename)                                                                                       
 96.00%  96.00%    9.11s     9.11s   __futex_abstimed_wait_common (/usr/lib64/libc.so.6)
 88.00%  88.00%    8.06s     8.06s   __lll_lock_wait (/usr/lib64/libc.so.6)
 40.00%  40.00%    3.52s     3.52s   pthread_cond_signal@@GLIBC_2.3.2 (/usr/lib64/libc.so.6)
 37.00%  37.00%    2.27s     2.27s   __lll_lock_wake (/usr/lib64/libc.so.6)
  4.00%  31.00%   0.640s     2.89s   pthread_mutex_lock@@GLIBC_2.2.5 (/usr/lib64/libc.so.6)
  3.00%  40.00%   0.490s     2.76s   __pthread_mutex_unlock_usercnt (/usr/lib64/libc.so.6)
  5.00% 266.00%   0.310s    24.40s   logging_log_callback (/project/src/av/logging.c)
 ```

The issue can be solved by actively disabling pyav's logging through `av.logging.set_level(None)`. 
However, since pyav by default hooks itself into python's logging system when `av.logging` is first imported,
there is no way to enforce this behavior if other modules are also using it (unless we import before everyone else, which isn't possible for library code).

I therefore propose two fixes:
1. Make logging async and lock free.
2. Do not hook into python's logging system unless explicitly asked.

PR #2394, opened 2026-08-25
Use a single GIL for logging

daveisfera:
I keep this one around for a while trying to decide if it was worth merging or not, because I worked around the issue in another way, but I think it's worth keeping. Basically, the concept is that if separate threads both log at the same time, then releasing the GIL for that copy can cause issues, so just keep it the whole time. I didn't do any benchmarks to see what the impact was, but I'm guessing that it's pretty minor for most users since they don't log much and releasing the GIL and then re-grabbing it has a cost of its own.

Looks like it's out of date with some changes you've made recently, so if you do want to keep this, then let me know and I'll rebase or feel free to pull it in or do it yourself.

And here's a repro that I tested the change with:
[repro_pyav_logging_segfault.py](https://github.com/user-attachments/files/31441786/repro_pyav_logging_segfault.py)

### DIFF
--- a/av/logging.py
+++ b/av/logging.py
@@ -331,25 +331,25 @@ def log_callback(
     if not inited:
         return
 
+    message: cython.char[1024]
+
     with cython.gil:
         if level > level_threshold and level != lib.AV_LOG_ERROR:
             return
 
-    # Format the message.
-    message: cython.char[1024]
-    vsnprintf(message, 1023, format, args)
+        # Format the message.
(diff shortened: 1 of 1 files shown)


WyattBlue:
Your repro is helpful. It's a good, specific bug report and worth its own issue. However, I must decline this patch. The nogil filter fast path addresses #2276, and the GIL doesn't serialize against the nogil teardown paths that cause the crash.

PR #2395, opened 2026-08-25
Accept GNU rpath flags from FFmpeg pkg-config.

daveisfera:
Source builds against a shared FFmpeg configured with --enable-rpath were failing on -Wl,-rpath, and even parsed -R flags never reached the linker.

`parse_cflags` was broken out to its own file to allow for testing


WyattBlue:
Just inline everything into `setup.py`, I don't want tests for `setup.py`
Other than that and fixing the conflicts, looks good.

PR #2401, opened 2026-08-28
riscv64: test smoke, tests workflows with RISE runners

threexc:


threexc:
My mistake, this is supposed to be on my fork. I'll re-open with results once I've fully tested.

PR #2407, opened 2026-09-02
PyAV: add riscv64 build support, use pyav-ffmpeg artifacts in smoke.yml

threexc:
Make use of RISE's Native RISC-V Runners to build PyAV for riscv64. This requires enabling them for the repository as a GitHub Application. More info regarding enabling them is available here: https://riscv-runners.riseproject.dev/

To get the `smoke.yml` workflow to succeed, I've reworked it to use pyav-ffmpeg artifacts like the `tests.yml` workflow does, except that the `armv7l` job still follows the existing approach. While working through this, I encountered a bug specific to the RISC-V Runners (`riscv64gc` profile), where I hit an illegal instruction/core dump issue that seems to be related to: https://www.mail-archive.com/ffmpeg-trac@avcodec.org/msg69006.html . Using the pyav-ffmpeg artifacts circumvented this issue, which is why it was implemented that way. This is a temporary workaround - in the future, the RISC-V Runners will support a more expansive profile. If using pyav-ffmpeg artifacts in the `smoke.yaml` workflow is undesirable, I can revisit it and maybe adjust compiler flags instead.

I tried a test run on my fork. You can review the results here: https://github.com/threexc/PyAV/pull/2

This is being done on behalf of [RISE](https://riseproject.dev/), using the [RISC-V Wheels](https://stanfromireland.github.io/riscv-wheels/) dashboard to track upstream support for the Python ecosystem.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -68,7 +68,7 @@ jobs:
         case ${{ matrix.config.os }} in
           ubuntu-24.04)
             sudo apt-get update
-            sudo apt-get install \
+            sudo apt-get install -y \
               autoconf \
               automake \
               build-essential \
@@ -79,7 +79,7 @@ jobs:
               pkg-config \
               zlib1g-dev
             if [[ "${{ matrix.config.extras }}" ]]; then
-              sudo apt-get install doxygen
+              sudo apt-get install -y doxygen
             fi
             ;;
           macos-14)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,6 +33,8 @@ jobs:
             arch: arm64
           - os: macos-15-intel
             arch: x86_64
+          - os: ubuntu-24.04-riscv
+            arch: riscv64
           - os: ubuntu-24.04-arm
             arch: aarch64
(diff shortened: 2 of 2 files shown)


threexc:
@WyattBlue the checks report says you cancelled the riscv64 ones. Is that just because the runners aren't enabled? Any general comments about this PR?

WyattBlue:
RISC runners are now enabled however:
 - There should only be one RISC job. No need to test doc building or old ffmpeg versions on new platforms.
 - macOS/Linux: ffmpeg should still be built from source.

threexc:
> RISC runners are now enabled however:
> 
>     * There should only be one RISC job. No need to test doc building or old ffmpeg versions on new platforms.
> 
>     * macOS/Linux: ffmpeg should still be built from source.

Alright. Thanks for enabling the runners. I'll experiment with some changes to the compile step so we don't re-use the pyav-ffmpeg archives.

threexc:
@WyattBlue what do you think?

WyattBlue:
Worthly enough to be worth using. A good follow up would be making the risv smoke tests faster, maybe by using pyav-ffmpeg binaries like Windows tests already does.

PR #2409, opened 2026-09-16
workflows: smoke: use pyav-ffmpeg for riscv64

threexc:


threexc:
Not bad - compared to the recent run on `main`, using pyav-ffmpeg saves ~27m in the riscv64 build.

PR #2410, opened 2026-09-17
Declare _get_hw_format nogil

adrianrfreedman:
`_get_hw_format` is installed as `AVCodecContext.get_format`, so libavcodec calls it from inside `avcodec_send_packet`, which PyAV runs under `with nogil` (`av/codec/context.pyx`). The callback therefore runs with no GIL held, and any Python object operation in it segfaults the interpreter.

It is already GIL-free in practice. `AVCodecPrivateData` holds only `AVPixelFormat` and `bint`, and since the pure-Python-mode migration the loop counter is typed (`i: cython.int = 0`), so nothing gets boxed. Nothing enforces that though. Declaring the callback `nogil` makes Cython reject a future change that reintroduces a Python object at compile time, rather than at runtime as a crash.

This is hardening, not a fix. The crash was real in 16.x, where the counter was untyped: `i += 1` compiled to `PyLong_FromLong`/`PyNumber_Add`, so every hwaccel initialisation failure took the process down with SIGSEGV instead of falling back to software decoding. 17.0.0 fixed it as a side effect of the typing change rather than deliberately, which is why I think the constraint is worth pinning. Background in #2411.

Verified on an L40S. 16.x segfaults on a stream NVDEC cannot decode (10-bit H.264), and this branch decodes the same file through the software fallback cleanly. Built with `python setup.py build_ext --inplace` against the vendored ffmpeg from `scripts/ffmpeg-latest.json` (libavcodec 63.1.101).

No functional change.


WyattBlue:
The commit message is so lousy, so I'd rather just add this myself.

adrianrfreedman:
> The commit message is so lousy, so I'd rather just add this myself.

This seems a bit unfair. Why not give me the opportunity to amend the commit?

